PR #127, opened 2024-01-22
feat(images): add steam bordered logo

eliobricenov:
https://www.figma.com/file/uKQtv5COIilU2x2dB5RpQH/Achievements-HyperPlay?node-id=1437:3260&mode=dev

I tried with

```css
    border: 5px solid #40FFBA;
    border-radius: 50%;
}
``` 

but it doesn't look quite as good, I think it is because the steam svg is not completely round. Either way I think we should't round it for all icons since we might want to non rounded icons

image


eliobricenov:
closing as this got stale, can revisit in the future

PR #278, opened 2024-05-15
feat(quest): add support for reward contract context

eliobricenov:
We decided to support linking the reward contracts to the projects. With this, we can reuse the same reward contract across quests.

With this PR we update the reward form card to support the following UX flow:

1. User selects network
2. Based on whether they have a contract available for the selected chain, we show the quest info inputs or the option to deploy a new contract

Figma: https://www.figma.com/design/KY1oPI1uvW2zOWr7RNt2JL/Dev-Portal?node-id=406-1040788&m=dev

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperplay/ui",
-  "version": "1.2.44",
+  "version": "1.2.45",
   "license": "UNLICENSED",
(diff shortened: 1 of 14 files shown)


eliobricenov:
closing as this got stale, can revisit in the future

PR #290, opened 2024-05-24
[Feat] Add overlay navbar component

BrettCleary:
# Summary

Design: https://www.figma.com/design/GKPELZxk551oVC3EDLgEDA/Current-Design?node-id=14764-20460&t=BFeIMlVO3i0r1ceF-4

# Storybook Link

https://hyperplay-ui-git-ui-addoverlaynavbar-hyperplay.vercel.app/?path=/docs/overlay-navbaroverlay--docs

# Screenshots
image


flavioislima:
a bit nitty-picky, but I find it weird that the items move only when you click on the Quests with the number on the side. Clicking on the other elements won't do that. Perhaps using a fixed height for the sidebar links could fix that 🤔 

https://github.com/user-attachments/assets/f9d716dc-9e99-42e7-a637-ab1c4197aadd

src/components/NavBarOverlay/index.tsx
flavioislima, line 1: i wonder if we should have a lib like `react-router` on the UI lib, it should not handle any routing on the app since it is not it's main function.

I think this component should receive a list of items as the props and those could be anything like a ReactNode and then it renders it with all the necessary props. 
This way we would not need add this responsability to the UI-lib since it should be only for the UI itself.

So on the Stories, instead of importing a whole HashRouter, perhaps only rendering some dumb elements would be ideal.

BrettCleary: great point. I just removed react router, decoupled the components like you said, and made NavItem polymorphic the same way `Button` is in Mantine https://mantine.dev/guides/next/#app--pages-router-together. So `NavItem` is agnostic to what router package/Link element is being used.

BrettCleary:
> a bit nitty-picky, but I find it weird that the items move only when you click on the Quests with the number on the side. Clicking on the other elements won't do that. Perhaps using a fixed height for the sidebar links could fix that 🤔
> 

Yup. It had a little bit of layout shift. I just set a min-height of 60px on `NavItem` to fix this and still keep it responsive.

flavioislima (approved):
LGTM, thanks refactoring the first version 💪🏼

PR #353, opened 2024-07-25
fix(quest-details): add min width to alert message icon

eliobricenov:


eliobricenov:
merging as this is minor change

PR #361, opened 2024-08-01
[feat]: Modify `QuestDetails` with latest design

red-game-dev:
- [x] Add cta component button child as prop
- [x] Reputation and Playstreak tags standalone

Paths: 

`/?path=/story/quests-questdetails--custom-cta-reputational-airdrop` and 
`/?path=/story/quests-questdetails--custom-cta-platstreak`


BrettCleary (approved):
nice. we'll need to decompose this component at some point 😅

PR #362, opened 2024-08-01
feat(check-email): add support for manual code input

eliobricenov:
Figma: https://www.figma.com/design/GKPELZxk551oVC3EDLgEDA/Current-Design?node-id=13220-35229&m=dev


eliobricenov:
merging as this is only used in a single place and design team gave the green light

PR #364, opened 2024-08-02
fix(collapse): make whole content clickable

eliobricenov:
Currently the button doesn't have any padding, the container does and as a result clicking the padding in the container doesn't trigger the collapse action.

By moving the padding to the button we can achieve making the padding space clickable.

Story: https://hyperplay-ui-git-fix-collapse-clickable-area-hyperplay.vercel.app/?path=/docs/collapse--docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperplay/ui",
-  "version": "1.6.7",
+  "version": "1.6.8",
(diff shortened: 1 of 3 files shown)


BrettCleary (approved):
nice catch! I think we have a similar issue on the account dropdown on the client

PR #369, opened 2024-08-02
[feat]: Added markdown to quest details

red-game-dev:
- [x] Introduce markdown on `QuestDetails`

**Note:** This wont update the markdown on `Collapse`, as there is plenty of places that would be affected at this stage. We do this on a different PR, as well we will introduce the changes on store, and other places that uses `Collapse`

Component Stories
- https://hyperplay-cwngs4aa1-hyperplay.vercel.app/?path=/story/quests-questdetails--markdown-description-custom-style
- https://hyperplay-cwngs4aa1-hyperplay.vercel.app/?path=/story/quests-questdetails--markdown-description


package.json
```diff
@@ -137,6 +138,7 @@
     "@hyperplay/utils": "^0.1.0",
     "@tabler/icons-react": "^2.46.0",
     "@valist/sdk": "^2.9.1",
-    "ethereum-blockies-base64": "^1.0.2"
+    "ethereum-blockies-base64": "^1.0.2",
+    "react-markdown": "^9.0.1"
```
BrettCleary, line 142: this needs to be added as a dev dependency

src/components/QuestDetails/index.tsx
```diff
@@ -199,13 +201,64 @@ export default function QuestDetails({
...
         <div className={cn('title', styles.title)}>{title}</div>
         <div className={cn('body-sm', 'color-neutral-400', styles.description)}>
-          {description}
+          <Markdown
+            {...markdownOptions}
+            components={markdownComponentsProp}
+            allowedElements={markdownAllowedElementsProp}
+          >
+            {description}
+          </Markdown>
```
BrettCleary, line 261: how do you feel about making this a separate component? let's call it `MarkdownDescription`

We can export that component from the ui library and pass it in as a `description` prop of type element to this Quest Details component

Then we can abstract it away and we have a reusable markdown component for the future

BrettCleary: Sidenote but I think we could do this for the other sections of the quest details too to better decompose this component

BrettCleary (changes requested):

src/components/QuestDetails/index.tsx
red-game-dev: Yup this can be done, certainly and I think it is better as it will be standalone component, generally just like the `Address` we discussed earlier

src/components/QuestDetails/index.tsx
```diff
@@ -206,9 +206,7 @@ export default function QuestDetails({
       <div className={cn(styles.container, classNames?.content)}>
         {sticker}
         <div className={cn('title', styles.title)}>{title}</div>
-        <div className={cn('body-sm', 'color-neutral-400', styles.description)}>
-          {description}
-        </div>
+        {description}
```
BrettCleary, line 209: is there a way we can preserve the above div styles if they only pass in a string for the description?

red-game-dev: Yup sorted, instead if we would need to style `MarkdownDescription`, we can just pass `classNames.root` and we can handle it. The functionality is reverted to the default container of `QuestDetails`

BrettCleary (approved):
LGTM

PR #370, opened 2024-08-03
[feat]: Allow QuestCard to accept div attributes

red-game-dev:
This is required as we will pass the `id` for scroll to specific quest on quests list page

### DIFF
--- a/src/components/QuestCard/index.tsx
+++ b/src/components/QuestCard/index.tsx
@@ -1,7 +1,10 @@
-import cn from 'classnames'
+import cn from 'classnames';
 
-import { CardGeneric, CardGenericProps } from '../CardGeneric'
-import styles from './index.module.scss'
+
+
+import { CardGeneric, CardGenericProps } from '../CardGeneric';
+import styles from './index.module.scss';
+
 
 export interface QuestCardProps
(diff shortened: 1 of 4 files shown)


src/components/QuestCard/index.tsx
```diff
@@ -44,7 +44,8 @@ export function QuestCard({
   classNames,
   selected,
   ...rest
-}: QuestCardProps) {
+}: QuestCardProps &
+  Omit<React.ComponentPropsWithoutRef<'div'>, keyof CardGenericProps>) {
```
BrettCleary, line 48: we should update the `QuestCardProps` type to include this

BrettCleary: nice job getting the div props + the mantine type. I missed that on some components

BrettCleary (approved):
just have 1 small suggestion with the types

src/components/QuestCard/index.tsx
red-game-dev: Updated, also FYI in case you would like to check, I've removed `className` as seems we are not using it at all, while we've got `className.root` which does the same functionality 

Effected Components:

`AchievementsCard`
`QuestCard`
`GenericCard`

PR #372, opened 2024-08-04
[fix]: Fixed QuestBanner overflow

red-game-dev:
The banner slide content was overflowing seems on certain resolutions


BrettCleary (approved):
I think it would be good to add a story at the resolution that was previously overflowing

PR #376, opened 2024-08-07
feat(check-email): add support for manual code input

eliobricenov:
Design: https://www.figma.com/design/GKPELZxk551oVC3EDLgEDA/Current-Design?node-id=13254-72194&m=dev


src/components/CheckEmail/index.tsx
```diff
@@ -66,21 +80,39 @@ const CheckEmail = ({
...
+        >
+          <TextInput
+            className={styles.input}
+            placeholder={i18n.inputPlaceHolder}
+            {...codeInputProps}
+          />
+        </Collapse>
+      </div>
       <div className={styles.linkContainer}>
         <span className={cn('button-sm', styles.subtitle)}>
```
BrettCleary, line 117: minor but I think this typography class is 'caption' on the latest designs

src/components/CheckEmail/index.tsx
```diff
@@ -66,21 +80,39 @@ const CheckEmail = ({
...
-          {i18n.subtitle} <span className="text--semibold">{email}</span>
+        <Modal.Body className={'body-sm'}>
+          {i18n.subtitle} <span className="text--semibold">{email}</span>.{' '}
+          {i18n.clickLoginTitleOr}{' '}
+          <Button
+            type="link"
+            className={styles.reEnter}
+            onClick={onReEnterEmail}
+          >
+            {i18n?.reEnterEmail}.
```
BrettCleary, line 95: Same comment as I left on another PR:
Breaking up all the strings will mess up how it gets translated. Using Trans should let you use components while keeping the translation string together https://react.i18next.com/latest/trans-component 

We make the same mistake elsewhere but I'm trying to call it out when I see it

BrettCleary (approved):
LGTM. just a couple minor comments 🙏

src/components/CheckEmail/index.tsx
eliobricenov: nice catch, I'd completely forgotten I did the same here

eliobricenov: We don't have i18n next in the repo though 😅

eliobricenov: what do you think about creating a ticket for this?

eliobricenov: added here https://github.com/HyperPlay-Gaming/hyperplay-ui/issues/384

PR #378, opened 2024-08-09
[feat]: Added View Reward and Claimed

red-game-dev:
Introduced new features for `QuestDetails`

- View Reward incl animation
- Claimed


src/components/QuestDetails/components/Reward/index.tsx
```diff
@@ -40,8 +44,24 @@ export default function Reward({ reward, i18n }: RewardProps) {
...
+
+    window.open(reward.marketplaceUrl, '_blank')
+  }
+
   return (
     <div key={reward.title} className={styles.rewardContainer}>
+      {reward.isClaimed ? (
+        <div className={styles.isClaimed}>{i18n.claimed}</div>
+      ) : null}
+      <div className={styles.viewRewardContainer} onClick={onViewRewardClick}>
```
BrettCleary, line 61: can you make this an a tag and set the src to `reward.marketplaceUrl`?

BrettCleary (approved):
looks good and is working. we need to make the view reward container an a tag though. Using the semantically correct tag allows users can copy paste the url, ctrl + click to open in a new tab, etc. which will be useful in the store

src/components/QuestDetails/components/Reward/index.tsx
red-game-dev: Yup can do 🙌

red-game-dev: Sorted 🙌

PR #383, opened 2024-08-10
[Tech] decouple search bar from quests summary table

BrettCleary:
# Summary

Decouple search bar from summary table.

Needed so that I can set `withinPortal` on the search bar for the store page scrolling, which is only needed on the store page.


red-game-dev (approved):
LGTM 🚀  Just one conflict to be solved 🙌

PR #387, opened 2024-09-12
fix(avatar): make online and offline mode same size

eliobricenov:
right now the offline avatar is 40x40 but the online avatar is 45x45


eliobricenov:
merging as I can confirm both avatars have the same dimension
image
image

PR #389, opened 2024-09-13
[tech] Always keep clear options button in view on options panel

jiyuu-jin:


src/components/OptionsAccordion/index.tsx
```diff
@@ -99,35 +99,34 @@ export default function OptionsAccordion({
...
+          <div className={styles.panelList}>{items}</div>
+          <Button
+            key={`${option}-clear-filter`}
+            type="tertiary"
+            size="small"
+            onClick={() => clearOptions(option)}
+            style={{
+              marginTop: 'var(--space-sm)',
+              border: '1.5px solid var(--color-stroke-01)'
+            }}
```
BrettCleary, line 117: can we move to a class

src/components/OptionsAccordion/index.tsx
```diff
@@ -99,35 +99,34 @@ export default function OptionsAccordion({
...
+            data-testid={`${option}-clear-filter`}
+          >
+            <div
+              className="button-sm"
+              style={{
+                color: 'var(--color-neutral-100)'
+              }}
```
BrettCleary, line 124:
```suggestion
              className="button-sm color-neutral-100"
```

let's try to avoid setting style explicitly and keep the styling coming from the css classes

src/components/OptionsAccordion/OptionsAccordion.module.scss
```diff
@@ -70,3 +70,8 @@ $containerRadius: var(--space-sm);
 .sectionTitle {
   padding: var(--space-sm) 0px;
 }
+
+.panelList {
+  max-height: 50vh;
```
BrettCleary, line 75: could we use something other than `vh`? `px`? I find that is usually breaks/is hard to work with

src/components/OptionsAccordion/index.tsx
jiyuu-jin: Yeah for sure, theses hard styles were already in the existing code but happy to refactor in this PR!

src/components/OptionsAccordion/OptionsAccordion.module.scss
jiyuu-jin: Rem?

BrettCleary (approved):
LGTM

PR #399, opened 2024-10-17
feat(quest-streak-progress): add support for right section

eliobricenov:
- Add support for passing a right section component. This will let us pass components conditionally
image
image

- Default to the playstreak bolts


BrettCleary (approved):
lgtm

PR #400, opened 2024-10-22
tech(quest-details): claim in quest detail

eliobricenov:
### Summary of changes:

- Remove claim all CTA and update rewards layout
image
image

- Added this version to the streak progress to support showing no percentage increase on non completed playstreaks.
image

- Changed title color for warning alert card
image


src/components/QuestDetails/components/Reward/index.tsx
```diff
@@ -4,50 +4,68 @@ import { decimalUnits, parseNumIntoReadableString } from '@hyperplay/utils'
...
   let numClaimsLeftComponent = null
+
   if (reward.numOfClaimsLeft) {
     const formattedNumOfClaimsLeft = parseNumIntoReadableString({
       num: reward.numOfClaimsLeft,
       units: decimalUnits,
-      minValue: '0.0001',
-      maxValue: '999999999'
+      minValue: '0',
+      maxValue: '9999'
```
eliobricenov, line 34: We don't have a lot of room in the layout, I decided to decrease the amount to 4 digits before formatting.

src/components/QuestDetails/index.tsx
```diff
@@ -81,7 +79,7 @@ export default function QuestDetails({
   let sticker = null
   let eligibilityReqComponent = null
   let buttonText = ''
-  let ctaClick = onClaimClick
+  let ctaClick = null
```
eliobricenov, line 82: CTA is not used by default; only use it for sync of sign-in.

src/components/QuestDetails/components/StreakProgress/index.tsx
```diff
@@ -58,6 +58,7 @@ function PlayStreakBolts({
 }
 
 export default function StreakProgress({
+  standby,
```
eliobricenov, line 61: Unfortunately, I couldn't prevent adding this optional prop to the streak progress. Currently, there is no way to show this version of the content. We need it because on the quest page, the play streak percentage shouldn't be increased.

image

src/components/QuestDetails/components/Reward/index.tsx
```diff
@@ -63,12 +81,33 @@ export default function Reward({ reward, i18n }: RewardProps) {
...
-        <div className={classNames(styles.titleContainer, 'menu')}>
-          <div className={styles.title}>{reward.title}</div>
-          {numToClaimComponent}
+      </div>
+      <div className={styles.infoContainer}>
+        <div className={styles.textContainer}>
+          <div className={classNames('title-sm', styles.title)}>
+            {formattedNumToClaim
+              ? `${formattedNumToClaim} ${reward.title}`
+              : reward.title}
```
eliobricenov, line 90: Currently, if formattedNumToClaim is null, we render the 'null' word

src/components/QuestDetails/components/StreakProgress/index.tsx
BrettCleary: this is fine, but I think we could also use this prop to prevent the update function from being called. Then the user could see their progress on the quest page in the client

src/components/QuestDetails/components/Reward/Reward.stories.tsx
```diff
@@ -0,0 +1,82 @@
+import type { Meta, StoryObj } from '@storybook/react'
+
+import RewardImage from '@/assets/RewardImage.png'
+
+import { QuestReward } from '../../types'
+import Reward from './index'
+
+const meta: Meta<typeof Reward> = {
+  title: 'Quests/Reward',
```
BrettCleary, line 9: should we change this to `Quests/QuestDetails/StreakProgress`? same as StreakProgress since it is only used by the quest details

src/components/QuestDetails/components/Reward/Reward.stories.tsx
```diff
@@ -0,0 +1,82 @@
...
+    reward: {
+      ...defaultReward,
+      numOfClaimsLeft: '1000000',
+      numToClaim: '1000'
+    },
+    i18n: defaultI18n
+  }
+}
+
+export const Loading: Story = {
```
BrettCleary, line 80: I don't see any difference between this story and `Default`

BrettCleary (changes requested):

src/components/QuestDetails/components/StreakProgress/index.tsx
eliobricenov: This is a great call; with that, this prop is not needed, thanks!

src/components/QuestDetails/components/Reward/Reward.stories.tsx
eliobricenov: yeah, there isn't, I'll remove this

src/components/QuestDetails/components/Reward/index.module.scss
```diff
@@ -108,8 +127,8 @@ $imgSize: 160px;
   align-items: center;
   position: absolute;
   text-transform: uppercase;
-  left: -43px;
-  top: 34px;
+  left: -50px;
+  top: 20px;
```
BrettCleary, line 131: can you add a todo to fix this later so it is dynamically positioned

BrettCleary (approved):
LGTM

PR #404, opened 2024-10-24
[UI] add navbar sublinks

BrettCleary:
# Summary
Nav bar sublinks with collapsible parent NavItem

prereq for https://github.com/HyperPlay-Gaming/product-management/issues/667

design here: https://www.figma.com/design/GKPELZxk551oVC3EDLgEDA/Overlay?node-id=16873-11201&t=oWG3rn8hFBa6Hkwg-4

# Testing
try it out here https://hyperplay-ui-git-ui-addnavbarsublink-hyperplay.vercel.app/?path=/docs/overlay-navbaroverlay--docs

# Screenshots

image

### DIFF
--- a/src/components/NavBarOverlay/NavBarOverlay.stories.tsx
+++ b/src/components/NavBarOverlay/NavBarOverlay.stories.tsx
@@ -2,7 +2,7 @@ import { useState } from 'react'
 
 import type { Meta, StoryObj } from '@storybook/react'
 
-import { Home, QuestIcon, TrophyOutline } from '@/assets/images'
+import { Home, MetaMask, QuestIcon, TrophyOutline } from '@/assets/images'
 
 import { NavBarOverlay, NavBarOverlayProps } from '.'
 import { NavItem } from '../NavItem'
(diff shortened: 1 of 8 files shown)


flavioislima:
quick question, `Marketplace` should not be in the selected status when any sub-link was selected?

BrettCleary:
> quick question, `Marketplace` should not be in the selected status when any sub-link was selected?

good point. let me fix this

pimentel08:
lgtm!

src/components/NavItem/SubLink/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { createPolymorphicComponent } from '@mantine/core'
+import classNames from 'classnames'
+
+import styles from './SubLink.module.scss'
+
+export interface SubLinkProps {
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  component?: any
+  /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
+  children?: any
```
eliobricenov, line 12: these can be ReactNode

src/components/NavItem/index.tsx
```diff
@@ -43,31 +52,85 @@ function _NavItem({
...
+  if (hasSublinks) {
+    const dropdownArrowClass: Record<string, boolean> = {}
+    dropdownArrowClass[styles.subMenuCollapsed] = subLinksCollapsed
+    dropdownArrow = (
+      <button
+        className={cn(styles.submenuCollapseIcon, dropdownArrowClass)}
+        onClick={(ev) => {
+          setSubLinksCollapsed?.(!subLinksCollapsed)
+          ev.preventDefault()
+          ev.stopPropagation()
```
eliobricenov, line 72: why did you need this?

src/components/NavBarOverlay/NavBarOverlay.stories.tsx
```diff
@@ -25,39 +26,106 @@ export const Default: Story = {
...
         key={'/achievements'}
-        onClick={() => setSelectedRoute('/achievements')}
+        onClick={() => console.log('coming soon')}
         collapsed={collapsed}
-        currentRoute={selectedRoute}
         secondaryTag={'Coming Soon'}
         style={{
-          cursor: 'not-allowed'
+          cursor: 'not-allowed',
+          opacity: '50%'
```
eliobricenov, line 126: let's take the opportunity to remove the inline style

BrettCleary: this is just a story though

src/components/NavItem/index.tsx
BrettCleary: I wanted to be able to support clicking the nav item to navigate when it has sublinks.

So if the arrow to collapse/expand the sublinks is clicked without this, it will expand/collapse but also navigate, which seems like bad ux.

For instance on the portfolio, the user could click the top level nav item and go to the portfolio, but they could also expand/collapse the sub items without navigating and then click a sub item to navigate.

Currently the way it is implemented on the client isn't using this and I'm just using the parent nav item to expand/collapse the sublinks as well, but I think it's nice that the component is flexible in this way

src/components/NavItem/SubLink/index.tsx
BrettCleary: updated children, but for component, Mantine uses `any` https://github.com/mantinedev/mantine/blob/fbcee929e0b11782092f48c1e7af2a1d1c878823/packages/%40mantine/core/src/core/Box/Box.tsx#L65. I think it is probably fine to just match them since we already know that works with next/link, react router, etc.

src/components/NavBarOverlay/NavBarOverlay.stories.tsx
eliobricenov: ahhhh never mind then

src/components/NavItem/index.tsx
eliobricenov: Understood. do you mind adding a one line explanation? I find it hard to understand the reasoning behind thosw two lines without this context

eliobricenov (approved):

src/components/NavItem/index.tsx
BrettCleary: sure thing. just added it

PR #405, opened 2024-10-27
[UI] Improvements to DownloadToast

flavioislima:


src/components/DownloadToast/utils.ts
```diff
@@ -0,0 +1,5 @@
+import * as fileSize from 'filesize'
+
+export const size = fileSize.partial({ base: 10, round: 1, spacer: '' }) as (
+  arg: unknown
+) => string
```
BrettCleary, line 5: Can we use https://github.com/HyperPlay-Gaming/utils/blob/main/src/parseNumIntoReadableString.ts instead? It is the same function we use everywhere else, it has full test coverage, and doesn't introduce another dependency.

flavioislima: We use `filesize` on the client already which won't duplicate the dependency when using there.
Another idea would be to avoid any calculation on this component and make it deal with the UI only, I think that should be better since those UI component should not really deal with any logic, only receive the info and render it.
What do you think?

flavioislima: done, changed for the utils method

src/components/DownloadToast/index.tsx
```diff
@@ -32,30 +33,15 @@ function getETAStringFromMs(etaInMs: number) {
...
+  if (days > 0) {
+    return `${days}d`
+  } else if (hours > 0) {
+    return `${hours}h:${minutes}m`
+  } else if (minutes > 0) {
+    return `${minutes}m:${seconds}s`
   } else {
-    return `0MB`
+    return `${seconds}s`
   }
```
BrettCleary, line 45: Would be nice if we could put this in `@hyperplay/utils` but completely understand if you don't have time right now. I think there are a couple other places we show an ETA string on the frontend and we could reuse that function in those components as well.

BrettCleary (approved):
LGTM!

src/components/DownloadToast/index.tsx
flavioislima: Just moved this method there and added a couple of others we use in the client.
Wrote tests for all of them as well.

PR #406, opened 2024-10-27
[UI] GameCard Patching Support

flavioislima:
Add Patching Statuses and actions.
Also removed the Pause button for the extracting method.


BrettCleary (approved):
LGTM

PR #414, opened 2024-11-12
UI/add to library button states

biliesilva:
Modify add/added to Library icon button

These are the changes: 

Game card state - not added to Library
image

Game card state - added to Library
image

[Design Link](https://www.figma.com/design/akS9JZprBjVCMk2onpPEcD/Store?node-id=159-40467&t=OAKbLD5FE8BiTiN7-1)


src/components/GameCard/index.tsx
biliesilva, line 1: I had to force the style on the popover because the Mantine light theme was overwriting the style applied in the CSS.

BrettCleary: I think you need to set the class under `classNames` for the top level `Popover` component. 

something like this if you rename your `styles.popover` class to `styles.dropdown`

```tsx
        <Popover
          position={pos}
          shadow="md"
          opened={showPopover}
          offset={offset}
          classNames={{ dropdown: styles.dropdown }}
        >
```

As a general rule, we favor `!important` in the css over explicitly setting style as well

BrettCleary: we can also make the popover `unstyled` and use a similar pattern as the AccountSelect https://github.com/HyperPlay-Gaming/hyperplay-ui/blob/89552c04fc714a011aeb77d41e7384cf5a382748/src/components/AccountSelect/index.tsx#L48-L54

You could copy the `popoverDropdown` css class over

BrettCleary:
Should we add an on hover effect here? Usually our buttons have a default opacity change on hover as the default behavior but I'm open to a different approach
image

src/components/GameCard/index.tsx
biliesilva: Changes made 😉

src/components/GameCard/index.tsx
```diff
@@ -295,15 +303,18 @@ const GameCard = ({
...
               onMouseLeave={close}
               style={!enableRemoveButton ? { opacity: '100%' } : {}}
             >
               <CheckmarkCircleOutline />
             </button>
           </Popover.Target>
-          <Popover.Dropdown className={styles.popover}>
+          <Popover.Dropdown
+            className={styles.popover}
+            style={{ backgroundColor: 'var(--color-neutral-700)' }}
```
flavioislima, line 316: Same here, you can remove it.

biliesilva: Done

flavioislima (approved):
Congratulations on your first PR 🥇 
Tested here and everything looks good now.
Good stuff!

BrettCleary (approved):
LGTM! great job :fire:  :partying_face:

PR #431, opened 2025-02-10
[UI] Add Quests and Developer links for Store consumption

ry-animal:
Screenshot 2025-02-10 at 3 57 13 PM

Screenshot 2025-02-10 at 3 58 02 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperplay/ui",
-  "version": "1.9.2",
+  "version": "1.9.3",
   "license": "LGPL-3.0-only",
   "scripts": {
     "dev": "vite",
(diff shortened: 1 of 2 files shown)


BrettCleary (approved):
lgtm. might need to update the breakpoints in the consuming code.

Also would be good if we could refactor this component to be more general. i.e. take the url's as props and the link text as props as well so we can support translations. We could make it take a list of nav item components so we wouldn't need to send ui lib pr's for small stuff like this

ry-animal:
> lgtm. might need to update the breakpoints in the consuming code.
> 
> Also would be good if we could refactor this component to be more general. i.e. take the url's as props and the link text as props as well so we can support translations. We could make it take a list of nav item components so we wouldn't need to send ui lib pr's for small stuff like this


I figured this would need some refactoring as a follow up - esp to be able to use with `website` since currently implementing it there would be a regression since website supports i18n also yes, making it generic would be ideal so dont't have to update -ui, then version update deps in the downstream.

biliesilva (approved):
LGTM, the only change that would be a plus is what Brett mentioned, updating breakpoints, also taking in consideration the distance between the menu and the logo for small screen (give it more breath space).

biliesilva (approved):

PR #434, opened 2025-02-13
[UI/UX] Add a back button to QuestDetails

ry-animal:
Screenshot 2025-02-12 at 11 21 23 PM

added www. to developers link to match playwright

### DIFF
--- a/src/components/NavBar/index.tsx
+++ b/src/components/NavBar/index.tsx
@@ -39,7 +39,7 @@ const NavBar = function ({ UserAvatar }: Props) {
       </a>
       <a
         className={`${navBarStyles.navItem} menu`}
-        href="https://hyperplay.xyz/developers"
+        href="https://www.hyperplay.xyz/developers"
         target="_blank"
         rel="noopener noreferrer"
         data-testid="developers"
--- a/src/components/QuestDetails/index.module.scss
+++ b/src/components/QuestDetails/index.module.scss
@@ -8,7 +8,7 @@
 .badges {
   display: flex;
   flex-direction: row;
-  align-items: start;
+  align-items: center;
   gap: var(--space-xs);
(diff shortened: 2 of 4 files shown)


src/components/QuestDetails/index.module.scss
```diff
@@ -59,3 +59,13 @@
...
 }
+
+.backButton {
+  width: var(--space-xl-fixed);
+  height: var(--space-xl-fixed);
+  min-width: var(--space-xl-fixed);
+  min-height: var(--space-xl-fixed);
+  padding: var(--space-xs);
+  border-radius: 100%;
+  border: 1px solid var(--color-neutral-100);
```
BrettCleary, line 70: let's use the default color styles for circular button for consistency across the app

src/components/QuestDetails/index.tsx
BrettCleary, line 1: this component is used in 4 different places across the app: 
1. desktop client main window quests tab
2. desktop client overlay
3. browser platform overlay
4. store quests page

so any changes need to be general enough for all of those use cases. We'll need a way to hide the circular button for 1-3

src/components/NavBar/index.tsx
```diff
@@ -39,7 +39,7 @@ const NavBar = function ({ UserAvatar }: Props) {
       </a>
       <a
         className={`${navBarStyles.navItem} menu`}
-        href="https://hyperplay.xyz/developers"
+        href="https://www.hyperplay.xyz/developers"
```
BrettCleary, line 42: this should be a separate PR so it doesn't get blocked by other changes you are introducing here

BrettCleary (changes requested):
Do we need a back button? On mobile, users should be able to use the back button of their browser right?

PR #435, opened 2025-02-13
[UI] Remove title rendering and make title optional

biliesilva:
## Description

This pr removes the Game Title from the carousel in the Storybook, the next step is to update it in the client. 

### Before
image

### After
image

### DIFF
--- a/src/components/Carousel/index.tsx
+++ b/src/components/Carousel/index.tsx
@@ -8,7 +8,7 @@ import Controller from './components/Controller'
 import styles from './index.module.scss'
 
 export interface SlideData {
-  title: string
+  title?: string
   slideElement: JSX.Element
   thumbnail?: JSX.Element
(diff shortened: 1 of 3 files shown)


src/components/Carousel/index.tsx
```diff
@@ -8,7 +8,7 @@ import Controller from './components/Controller'
 import styles from './index.module.scss'
 
 export interface SlideData {
-  title: string
```
BrettCleary, line 11: can we just remove this prop? is it being used anywhere in this component?

BrettCleary (approved):
lgtm

ry-animal (approved):

src/components/Carousel/index.tsx
biliesilva: Removed

PR #439, opened 2025-02-20
[Fix] Streak progress when not playing

BrettCleary:
# Summary

add `streakIsProgressing` prop to enable the % completion interpolation which was on by default previously


ry-animal (approved):
sick!

eliobricenov (approved):

flavioislima (approved):

PR #440, opened 2025-02-21
[feat] Add game name to game page

biliesilva:
## Description

This PR is the first step to adding the game title to a location other than the carousel, as requested by @jacobc-eth. The game title was previously removed from the carousel component as asked, resulting in no game name on the game details page. Based on feedback from @pimentel08, we decided to put the game title in the About section, as it will be the future side place for that info. 

### Future desired state
image

### Temporary Change
image

### DIFF
--- a/src/components/GameAbout/GameAbout.module.scss
+++ b/src/components/GameAbout/GameAbout.module.scss
@@ -28,9 +28,5 @@
 
 .gameTitle {
   padding-bottom: var(--space-md);
-  h2 {
-    margin: 0;
-    color: var(--color-neutral-200);
-    font-size: var(--text-6xl);
-  }
+  color: var(--color-neutral-200);
 }
--- a/src/components/GameAbout/GameAbout.module.scss
+++ b/src/components/GameAbout/GameAbout.module.scss
@@ -25,3 +25,12 @@
     height: 16px;
   }
 }
+
+.gameTitle {
+  padding-bottom: var(--space-md);
+  h2 {
+    margin: 0;
+    color: var(--color-neutral-200);
+    font-size: var(--text-6xl);
+  }
+}
--- a/src/components/GameAbout/index.tsx
+++ b/src/components/GameAbout/index.tsx
@@ -33,10 +33,14 @@ const GameAbout = ({ description, gameName, classnames }: GameAboutProps) => {
   return (
     <div className={classnames?.container}>
       {gameName && (
-        <div className={styles.gameTitle}>
-          <h2 className={classNames('h4', styles.title, classnames?.gameTitle)}>
-            {gameName}
-          </h2>
+        <div
+          className={classNames(
+            'header',
+            styles.gameTitle,
+            classnames?.gameTitle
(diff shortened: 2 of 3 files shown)


flavioislima (approved):
Good stuff 🔥

src/components/GameAbout/GameAbout.module.scss
```diff
@@ -25,3 +25,12 @@
...
   }
 }
+
+.gameTitle {
+  padding-bottom: var(--space-md);
+  h2 {
+    margin: 0;
+    color: var(--color-neutral-200);
+    font-size: var(--text-6xl);
+  }
```
BrettCleary, line 35: Is it possible for us to use a type from our typography system?

src/components/GameAbout/index.tsx
```diff
@@ -26,6 +32,13 @@ const GameAbout = ({ description, classnames }: GameAboutProps) => {
 
   return (
     <div className={classnames?.container}>
+      {gameName && (
+        <div className={styles.gameTitle}>
+          <h2 className={classNames('h4', styles.title, classnames?.gameTitle)}>
```
BrettCleary, line 37: could we just make this an h4 element instead of applying an h4 class to an h2 element?

BrettCleary:
Is there a design for this? The title looks too large

biliesilva:
> Is there a design for this? The title looks too large

Not exactly, I talked with pimentel about the place and used this design for our future store page I shared above, and there we are using an h6, but I think it is different from the way we are using variables in the code now, so I tried to put the variable that could be more similar to this design.

src/components/GameAbout/GameAbout.module.scss
biliesilva: I think I forgot to define this here, because when I put the mock in stories it used the same typography we already used: Poppins. 🤔

 I will fix this, thanks for the tip

src/components/GameAbout/index.tsx
```diff
@@ -26,6 +32,17 @@ const GameAbout = ({ description, classnames }: GameAboutProps) => {
 
   return (
     <div className={classnames?.container}>
+      {gameName && (
+        <div
```
eliobricenov, line 36: Maybe we should use a header tag instead of a div.

eliobricenov (approved):
code lgtm

src/components/GameAbout/index.tsx
BrettCleary: for this one elio meant the tag should be one of `h1`, `h2`, `h3`... `h6`. The `header` tag is meant for a different use case.

From https://www.w3schools.com/tags/tag_header.asp
>A `<header>` element typically contains:
> one or more heading elements `(<h1> - <h6>)`
> logo or icon
> authorship information


If we want to use the `header` css class, then it could still be one of `h1-h6`, but by specifying the `header` css class, it would override the default styling while keeping it semantic.

In general, we aren't very semantic though so I'm fine with it being a div

PR #445, opened 2025-02-28
(fix) fix styling of the quest card when showing rewards

biliesilva:
## Description
- This pr is to fix the styling of the reward section in the quest cards for the store. Right now, this section seems broke,n also giving a bad ux experience to the player who wants to be informed about the reward itself; 
- The next step is to bring this change into the store repository.

### How it is now
image

### Changes
image


flavioislima (approved):
Good stuff!

biliesilva:
@BrettCleary could you check this pr?

src/components/QuestCard/index.tsx
```diff
@@ -92,32 +92,31 @@ export function QuestCard({
...
+                />
+              </div>
               <div
                 className={cn(styles.currencyName, classNames?.currencyName)}
               >
                 {currencyName}
               </div>
             </div>
+
+            {currenyAmount && (
```
BrettCleary, line 115:
```suggestion
            {currencyAmount && (
```

BrettCleary: should go ahead and fix this mispelling

src/components/QuestCard/index.module.scss
```diff
@@ -51,44 +51,59 @@ $border: 2px var(--color-neutral-600) solid;
...
+  font-size: var(--font-size-sm);
+  color: var(--color-success-100);
+  font-weight: var(--semibold);
+}
+
 .questName {
   color: var(--color-neutral-100);
   font-weight: var(--medium);
+  font-size: var(--text-lg);
+  font-weight: var(--semibold);
```
BrettCleary, line 82: font-weight is specified twice on this class

src/components/QuestCard/index.module.scss
BrettCleary, line 1: would be good to reuse existing typography types :+1: unless the design is specifically deviating from our typography system :fearful:

BrettCleary (approved):
LGTM. I think we should add a story with one of these odd sized cover images 
image

since that is the edge case we missed previously

PR #450, opened 2025-03-06
feat(NativePlatformInput): Add i18n support for upload translations

eliobricenov:


src/components/NativePlatformInput/index.tsx
```diff
@@ -50,6 +50,7 @@ export function NativePlatformInput({
       }}
       exeInputProps={exeInputProps.amd64}
       exeList={exeListAmd64}
+      i18n={i18n?.upload}
```
eliobricenov, line 53: I'll merge since this is the only change

PR #455, opened 2025-03-12
[CI] Add workflow to check for version bump in package.json

flavioislima:
Adds a new workflow that will warn if the version in package.json was not changed on the PR since this happens to us from time to time and we then need another commit to change it.

### DIFF
--- a/.github/workflows/check-version-bump.yml
+++ b/.github/workflows/check-version-bump.yml
@@ -1,4 +1,4 @@
-name: Check Package.json Version Bump
+name: Check Package.json Version Bump (Informational)
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


flavioislima:
For some reason I am not being able to write a reusable workflow for this, opened this second PR with the change and added the workflow to the workflows repository but for some reason it doesnt trigger.

Would be good since we probably want this on the other libs as well.

https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/457
https://github.com/HyperPlay-Gaming/workflows/commit/721063c4082eedeebd59271986b1a4134d6c955c

BrettCleary (approved):

PR #459, opened 2025-03-13
[Feat] Update and refactor carousel 

BrettCleary:
# Summary

closes https://github.com/HyperPlay-Gaming/px-pm/issues/128

This refactors the existing carousel to be more modular and easier to work with in consuming code.

Instead of prop drilling, we now match the same architecture of Mantine effectively with `Carousel` being a provider and all other `Carousel.XXX` components using this provider.

Subcomponents of components were moved into `components` folders at the same level as that component. 

The controller is now `Carousel.Controller` and slides are `Carousel.Slides` to decouple the components but still indicate that they are subcomponents, similar to mantine.

Storybook tests exist at the Carousel level for all major user flows, the tests with `skip-test` tag are unable to run in CI for one reason or another, but pass locally.

This updates to the latest carousel design and updates several outdated dependencies as well.

Storybook viewports are now supported as well, which was necessary for mobile testing, and overall improves the storybook viewing experience.


biliesilva (approved):
The mobile version is expected to have the controller pics big like this: 
image

Some tests are also failing

BrettCleary:
> Is the mobile version expected to have the controller pics big like this below? 
> 
> Some tests are also failing 
> 
> 

Yes. This should match the design. 

Which tests are failing?

src/components/Carousel/components/Controller/components/Item/Item.module.scss
```diff
@@ -0,0 +1,151 @@
+@use 'sass:math';
+@use '../../../../../../styles/utilities/variables' as *;
```
flavioislima, line 2: cant these be relative paths?

src/components/Carousel/components/Controller/components/BaseButton/index.tsx
flavioislima, line 1: Could this button maybe be on the main components folder? it looks generic enough so we could use it in other places.
For instance, it looks similar to the one we will have on the new Story sections to navigate horizontally.
Wdyt?

flavioislima (approved):
Left a few comments but otherwise looks good.

src/components/Carousel/components/Controller/components/Item/Item.module.scss
BrettCleary: I think you meant to say absolute path, and I haven't been able to get absolute paths working on sass imports yet. We should revisit on another pr as it will likely require some vite config or storybook config changes

src/components/Carousel/components/Controller/components/BaseButton/index.tsx
BrettCleary: great point. I just moved it and renamed it to `ArrowCircularButton`

PR #461, opened 2025-03-13
[Fix] Edge case where downloaded size is bigger than download size

flavioislima:
This adds a safeguard to the Download Toast so it won't show a Downloaded size bigger than Total Download Size.
This happens to a few HyperPlay games while extracting.

image


flavioislima:
Will try to solve this on the client first

PR #463, opened 2025-03-18
[Feat] New GameInfoV2 component

biliesilva:
# Component Features
- Responsive game card layout with cover image and game information
- Support for game metadata display:
  - Game title and version
  - Badges
  - Blockchain component
  - Social media links component
  - Editor's Choice indicator

# New components
- BlockchainStack
- GameInfoV2
- SocialLinks

# Technical Details
- Added Storybook test for the onHover state for the BlochchainStack component 
- Added i18n support 

## Design
image

## Storybook
image

### Mobile 
image


flavioislima:
Code looks good. 
Only thing I found was that, maybe its by design, but without the social links it looks a bit broken:
image

flavioislima (approved):
Good stuff, looks pretty! 😍

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+const BlockchainIcon: React.FC<{
+  blockchain: BlockchainIconProps
+  className?: string
+}> = ({ blockchain, className }) => (
+  <div className={classNames(styles.blockchainIcon, className)}>
+    {React.createElement(blockchain.icon, {
+      className: styles.blockchainSvg
+    })}
+  </div>
+)
```
BrettCleary, line 110: these should be in a `components` folder as separate react components

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+const MAX_VISIBLE_BLOCKCHAINS = 5
```
BrettCleary, line 112: can we just take this as an optional prop with 5 as the default value?

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+          <div className={styles.socialLinks}>
+            {socialLinks?.map((link, index) => (
+              <Button
+                key={index}
+                type="secondary"
+                size="icon"
+                onClick={() => window.open(link.url, '_blank')}
+              >
+                <link.IconButton className={styles.socialIcon} />
+              </Button>
```
BrettCleary, line 220: this should be a separate component and we should use a similar structure as on the `gamePage` in the store. It should take a Link component and have a default value as an `a` tag. You can see an example for the `Link` prop in `src/components/NavBar/index.tsx`

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+                  Early Access
+                </Sticker>
+              ) : null}
+              <Sticker styleType="neutral" variant="outlined">
+                {title}
+              </Sticker>
+              {playerCount ? (
+                <Sticker styleType="neutral" variant="outlined">
+                  {playerCount}
+                </Sticker>
```
BrettCleary, line 202: left/right padding is off here 8px on the preview and 12px on the design.

image

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+          ))}
+          {showRemainingCount && (
+            <div className={styles.more}>
+              <span className={styles.moreCount}>+{remainingCount}</span>
+            </div>
+          )}
+        </div>
+      </div>
+    )
+  }
```
BrettCleary, line 162: let's make this a separate component

BrettCleary: also the on hover styling/interactions don't seem to match the design. can we check with design here? I think the desired behavior was only `Popover` on the `+X` count which would show a list of the remaining chains

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  const renderBlockchains = () => {
+    if (!blockchains || blockchains.length === 0) return null
+
+    const visibleBlockchains = blockchains.slice(0, MAX_VISIBLE_BLOCKCHAINS)
+
+    return (
+      <div className={styles.blockchains}>
+        <span className={styles.blockchainLabel}>
+          {i18n?.blockchainLabel || 'BLOCKCHAIN(S):'}
```
BrettCleary, line 148: you can avoid this by providing a default initialization to the i18n variable above. For instance, this is how we do it on the `NavBar`

```
i18n = {
    installHyperPlay: 'Install HyperPlay'
  }
```

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  const renderEditorChoice = () => {
+    if (!editorChoice?.isEditorChoice) return null
+
+    return (
+      <div className={classNames(styles.editorChoice, editorChoice.className)}>
+        <Images.EditorChoice />
+        Editor&apos;s Choice {editorChoice.year || new Date().getFullYear()}
+      </div>
+    )
+  }
```
BrettCleary, line 138:
```suggestion
  let editorChoice = null
  if (editorChoice?.isEditorChoice){
    editorChoice = (
      <div className={classNames(styles.editorChoice, editorChoice.className)}>
        <Images.EditorChoice />
        Editor&apos;s Choice {editorChoice.year || new Date().getFullYear()}
      </div>
    )
  }
```

This is the way we usually do conditional rendering for components we don't want to break out for readability

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,231 @@
...
+  if (!src || hasError) {
+    return fallback ? (
+      <div className={classNames(styles.imageFallback, className)}>
+        {fallback}
+      </div>
+    ) : null
+  }
+
+  return (
+    <img
```
BrettCleary, line 92: we should be able to support `next/image` here instead of just `img`. I think we should take the image tag as a prop

src/assets/images/LoadingIcon2.svg
BrettCleary, line 1: where are we using this icon on this PR?

src/assets/images/X.svg
BrettCleary, line 1: we already have `src/assets/images/XLogo.svg` and `src/assets/images/XLogoFilled.svg`

src/assets/images/Plus.svg
BrettCleary, line 1: do we need to add this if we're just using it in the story?

src/assets/images/Polygon.svg
BrettCleary, line 1: hmm I think we should add `@hyperplay/chains` as a dev dependency and use those images for these chains in the story since these chain icons are only added for the story, which I think we should avoid.

src/assets/images/DiscordFilled.svg
```diff
@@ -1,12 +1,10 @@
...
+</g>
+<defs>
+<clipPath id="clip0_2704_43722">
+<rect width="24" height="24" fill="white"/>
```
BrettCleary, line 7: we need this to be `current` so that it is flexible and we can specify the color as a `fill` attribute on the icon. This way we don't have to export svg's for every single color we want to support, we can use a design system color, and we can do things like animate it or change the color onHover, etc.

src/assets/images/YoutubeFilled.svg
BrettCleary, line 1: can we use `src/assets/images/Youtube.svg`?

src/assets/images/EditorChoice.svg
```diff
@@ -0,0 +1,17 @@
...
+<path d="M12.495 9.39111C12.1174 10.5087 11.9286 13.6743 16.0648 15.2528C16.0648 13.3004 15.3359 10.4347 12.495 9.39111Z" fill="#9469FE" fill-opacity="0.75"/>
+<path d="M6.85242 14.6073C8.54729 16.2815 10.6154 16.4685 14.2115 15.1683C11.3662 13.1158 7.98087 13.8638 6.85242 14.6073Z" fill="#9469FE" fill-opacity="0.75"/>
+<path d="M7.60077 7.15967C7.22316 9.39477 9.4625 11.2298 11.2057 11.943C10.9686 9.39477 9.33955 7.67279 7.60077 7.15967Z" fill="#9469FE" fill-opacity="0.75"/>
+<path d="M4.03809 10.241C4.91626 11.8891 7.59907 12.2891 9.7989 12.5674C9.10881 11.8193 8.26627 11.2247 7.32695 10.8229C6.38763 10.4211 5.37291 10.2213 4.34984 10.2367L4.03809 10.241Z" fill="#9469FE" fill-opacity="0.75"/>
+<path d="M6.55833 9.71444C6.20267 5.94868 0.411124 5.65298 0.165235 0C-1.03347 7.39236 4.61757 8.86649 6.55833 9.71444Z" fill="#9469FE" fill-opacity="0.75"/>
+<path d="M19.8721 20.0001C20.3911 20.0001 20.8118 19.5834 20.8118 19.0695C20.8118 18.5555 20.3911 18.1389 19.8721 18.1389C19.3532 18.1389 18.9325 18.5555 18.9325 19.0695C18.9325 19.5834 19.3532 20.0001 19.8721 20.0001Z" fill="#9469FE" fill-opacity="0.75"/>
+</svg>
```
BrettCleary, line 17: can we change all the `fill`'s to `fill="current"` so we can set it dynamically on the component?

src/components/GameInfoV2/index.tsx
BrettCleary: We may want to consider making this a separate component if we'll use it elsewhere

BrettCleary (changes requested):
Sorry. Lots of change requests here to keep things modular and maintainable. 

After refactoring the components let's add tests for interactions like `moreCount` onHover and what not as well.

src/assets/images/X.svg
biliesilva: These are different logos, the one we have is outlined and in the design is filled.

src/components/GameInfoV2/index.tsx
biliesilva: I talked with @pimentel08 and she said we have a tooltip on the count showing icon and name of the rest of the blockchains.

src/components/GameInfoV2/index.tsx
biliesilva: I can put the link, but for a new component, I don't think it's necessary as I'm using the button component already,y which has a variant called "Icon Button".

src/assets/images/DiscordFilled.svg
biliesilva: They are current now, but even applying the fill in SVG path is not working, it makes the icons invisible.

src/assets/images/EditorChoice.svg
biliesilva: same as my comment above

src/components/GameInfoV2/index.tsx
biliesilva: This exactly design will live just here, checked with @iamdavid0091

src/components/GameInfoV2/index.tsx
biliesilva: I don't think it's possible to use next in the UI lib. Could you explain this?

src/assets/images/Plus.svg
biliesilva: Removed

src/assets/images/DiscordFilled.svg
biliesilva: Changed

src/assets/images/EditorChoice.svg
biliesilva: Changed

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+                  <Sticker
+                    key="playerCount"
+                    styleType="neutral"
+                    variant="filledStrong"
+                  >
+                    {playerCount}
+                  </Sticker>
+                ]}
+                classNames={{}}
+                maxVisibleItems={5}
```
BrettCleary, line 153: I think we should make this something like 20. The popover for the meta section component still needs some styling tweaks and design/prod mentioned to me that we should just show all the tags for now

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        className: styles.youtubeIcon
+      }
+    }
+  }
+
+  return (
+    <div className={classNames(styles.container, className)}>
+      <div className={styles.imageContainer}>
+        {isLoading ? <div className={styles.imageLoading} /> : ImageComponent}
+      </div>
```
BrettCleary, line 109: I believe we will hide the image on mobile 
image

I saw your comment on figma but I believe @pimentel08 was referring to the carousel instead of this game image

src/components/GameInfoV2/GameInfoV2.module.scss
```diff
@@ -0,0 +1,245 @@
+@use '../../styles/utilities/loading' as *;
+
+.container {
+  display: flex;
+  gap: var(--space-xl);
+  justify-content: center;
+  align-items: space-between;
+  align-self: stretch;
+  padding: var(--space-2xl) var(--space-3xl);
+  background-color: var(--background-base);
```
BrettCleary, line 10: Isn't the background transparent in the designs?

src/components/GameInfoV2/GameInfoV2.module.scss
```diff
@@ -0,0 +1,245 @@
+@use '../../styles/utilities/loading' as *;
+
+.container {
+  display: flex;
+  gap: var(--space-xl);
+  justify-content: center;
+  align-items: space-between;
+  align-self: stretch;
+  padding: var(--space-2xl) var(--space-3xl);
```
BrettCleary, line 9:
```suggestion
```

This is too opinionated I think and should be handled by the consuming code.

src/components/GameInfoV2/GameInfoV2.module.scss
```diff
@@ -0,0 +1,245 @@
...
+  background-color: var(--background-base);
+  width: 100%;
+
+  @media (max-width: 768px) {
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+    max-width: fit-content;
+    align-items: flex-start;
+    padding: var(--space-2md) var(--space-xs);
```
BrettCleary, line 19:
```suggestion
```

src/components/GameInfoV2/GameInfoV2.module.scss
```diff
@@ -0,0 +1,245 @@
...
+  width: 100%;
+  height: 100%;
+  object-fit: cover;
+}
+
+.imageLoading {
+  @include shinyLoader;
+  width: 100%;
+  height: 100%;
+  border-radius: var(--space-xs);
```
BrettCleary, line 97: could make this a sass variable to share with `.imageContainer`

src/components/GameInfoV2/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <div className={styles.actions}>
+          <div className={styles.socialLinks}>
+            {socialLinks?.map((link, index) => {
+              const { icon: Icon } = socialIcons(link.type) ?? {}
+              return (
+                <Button
+                  key={index}
+                  type="secondary"
+                  size="icon"
+                  onClick={() => window.open(link.url, '_blank')}
```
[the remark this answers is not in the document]
BrettCleary: I'd recommend this being a separate component that takes the social and returns this link button. I could see it being used elsewhere

BrettCleary (changes requested):
getting closer. still some issues that need to be addressed though

src/components/GameInfoV2/GameInfoV2.module.scss
biliesilva: I saw that it was changed too because before it was filled

src/components/BlockchainsStack/BlockchainsStack.module.scss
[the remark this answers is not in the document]
biliesilva: I put a background color for these icons because there are some icons that have transparency. this C and ethereum are ones of this, But I see it is not filling all the space too :(

BrettCleary (approved):
looks great! :fire:

PR #469, opened 2025-03-25
[Feat] Add new Rewards Cards and Rewards Section

flavioislima:
This pull request introduces several new components and enhancements to the existing components in the project. The key changes include the addition of the `RewardsCard` and `RewardsSection` components, updates to the `CardGeneric` component, and various improvements to the styling and functionality of the rewards-related components.

### New Components:
* **RewardsCard Component**:
  - Introduced a new `RewardsCard` component with customizable properties for displaying game rewards.
  - Added a Storybook story for the `RewardsCard` component to demonstrate its usage.
  - Included styles for the `RewardsCard` component to handle different states such as loading.

* **RewardsSection Component**:
  - Added a new `RewardsSection` component that displays a horizontally scrollable collection of reward cards with navigation controls.
  - Created a Storybook story for the `RewardsSection` component to showcase its features and different use cases.
  - Provided styles for the `RewardsSection` component to ensure responsive design and proper layout.

### Enhancements to Existing Components:
* **CardGeneric Component**:
  - Introduced a new prop `showGradientBorderAndShadow` to control the display of gradient border and shadow. [[1]](diffhunk://#diff-ab3f18a6265c5ac66883a73fbf86a09659dcb3a8be8f91a573c2123de574505fR31) [[2]](diffhunk://#diff-ab3f18a6265c5ac66883a73fbf86a09659dcb3a8be8f91a573c2123de574505fR51-R62)

### Miscellaneous Updates:
* **Footer Component**:
  - Added a new link to the cookie policy in the `Footer` component.

* **Export Updates**:
  - Updated the `src/index.ts` file to export the new `RewardsCard` and `RewardsSection` components along with their respective types.
  - Exported the `ArrowCircularButton` component for use in other parts of the application.

* **Version Update**:
  - Updated the package version in `package.json` from `1.16.2` to `1.17.0`.

### DIFF
--- a/src/components/RewardsSection/index.tsx
+++ b/src/components/RewardsSection/index.tsx
@@ -30,11 +30,30 @@ const RewardsSection = ({
     loop: false
   })
 
+  const [isScrollable, setIsScrollable] = useState(false)
+  const [timeoutReached, setTimeoutReached] = useState(false)
+
+  useEffect(() => {
+    let timeoutId: NodeJS.Timeout
+
+    if (isLoading) {
+      timeoutId = setTimeout(() => {
+        setTimeoutReached(true)
+      }, 5000)
+    } else {
+      setTimeoutReached(false)
+    }
+
+    return () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId)
+      }
+    }
+  }, [isLoading])
+
   if (isLoading) {
     rewards = dummyData
   }
-
-  const [isScrollable, setIsScrollable] = useState(false)
 
   const scrollPrev = useCallback(() => {
     if (!emblaApi) return
@@ -65,7 +84,7 @@ const RewardsSection = ({
     setIsScrollable(emblaApi.canScrollNext() || emblaApi.canScrollPrev())
   }, [emblaApi])
 
-  if (!rewards || rewards.length === 0) {
+  if (!rewards || rewards.length === 0 || (isLoading && timeoutReached)) {
     return null
   }
 
--- a/src/components/RewardsSection/RewardsSection.stories.tsx
+++ b/src/components/RewardsSection/RewardsSection.stories.tsx
@@ -226,8 +226,7 @@ export const FewItems: Story = {
   }
 }
 
-// not passing rewards so we can test the dummy data loading state
-export const LoadingWithDummyData: Story = {
+export const LoadingTimeout: Story = {
   args: {
     Link: LinkComponent,
     isLoading: true
@@ -239,9 +238,23 @@ export const LoadingWithDummyData: Story = {
     docs: {
       description: {
         story:
-          'Loading state is indicated by a spinner or skeleton loader on the reward cards, providing visual feedback to users while data is being fetched.'
+          'When loading state persists beyond 5 seconds, the component returns null. This story simulates the timeout behavior by forcing the loading state to continue past the timeout threshold.'
       }
     }
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+
+    // Check that the component is rendered initially with loading state
+    let rewardsSection = canvas.queryByTestId('rewards-section')
+    expect(rewardsSection).toBeInTheDocument()
+
+    // Wait for more than 5 seconds (5500ms to be safe)
+    await new Promise((resolve) => setTimeout(resolve, 5500))
+
+    // After 5 seconds, the component should return null
+    rewardsSection = canvas.queryByTestId('rewards-section')
+    expect(rewardsSection).not.toBeInTheDocument()
   }
 }
 


src/components/RewardsCard/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+      genericClassNames={{
+        body: styles.rewardsCard,
+        image: styles.rewardImage
+      }}
+      className={styles.cardBase}
+    >
+      <div className={styles.content}>
+        <div className={styles.stickers}>
+          <Sticker styleType="neutral" dimension="default">
+            {reward}
```
biliesilva, line 41: The design changed, and now the sticker is the one called "filledStrong" neutral. if you do not put the type here, it will load the first of the list, I believe. 
image

biliesilva (approved):

src/components/RewardsCard/index.module.scss
```diff
@@ -0,0 +1,49 @@
+.cardBase {
+  width: 271px;
+  // should be 379px when we update it to use two stickers
+  height: 339px;
```
BrettCleary, line 4: can we not hardcode the height and have it fit the content?

src/components/RewardsCard/index.module.scss
```diff
@@ -0,0 +1,49 @@
...
+}
+
+.content {
+  text-align: center;
+}
+
+.rewardImage {
+  position: relative;
+  width: 100%;
+  height: 271px;
```
BrettCleary, line 28: could we make this more dynamic than just hardcoding a height?

src/components/RewardsCard/index.module.scss
```diff
@@ -0,0 +1,49 @@
...
+  justify-content: space-between;
+  background-color: var(--color-neutral-600);
+  border-radius: var(--space-xs);
+}
+
+.rewardsCard {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  padding: 16px;
```
BrettCleary, line 15:
```suggestion
  padding: var(--space-md-fixed);
```

src/components/RewardsCard/index.tsx
BrettCleary, line 1: we need a loading state here as well since this will be a client side fetch. Needs an `isLoading` prop and can use the `shinyLoader` mixin for that state

src/components/CardGeneric/index.tsx
```diff
@@ -47,15 +48,16 @@ export function CardGeneric({
   children,
   className,
   genericClassNames,
+  isReward = false,
```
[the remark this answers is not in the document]
flavioislima: makes sense, will do

src/components/RewardsCard/index.tsx
flavioislima: Cool, thanks for the heads up

src/components/RewardsCard/RewardsCard.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+  },
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+
+    // Check that the reward text is displayed correctly
+    expect(canvas.getByText('1000 YGG points')).toBeInTheDocument()
+
+    // Check that the card renders without errors
+    const stickers = canvas.getAllByRole('generic', { hidden: true })
+    expect(stickers.length).toBeGreaterThan(0)
```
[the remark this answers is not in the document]
flavioislima: They were not duplicates until I removed half of them due to removing the claimLeft stuff.
Will update those

src/components/RewardsCard/RewardsCard.stories.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const imgElement = canvas.getByRole('img')
+    expect(imgElement).toBeInTheDocument()
+    expect(imgElement).toHaveAttribute(
+      'src',
+      expect.stringContaining('xocietyNTx')
+    )
+  }
+}
+
+export const CustomLabels: Story = {
```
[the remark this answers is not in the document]
flavioislima: To test if the translations are being passed on

src/components/RewardsCard/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+function RewardsCard({
+  rewardImage,
+  reward,
+  claimsLeft,
+  i18n = {
+    claimsLabel: 'Claims left',
+    claimsLeftLabel: 'Unlimited'
```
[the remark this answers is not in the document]
flavioislima: Hmm, it was on the design though.

src/components/RewardsCard/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <div className={styles.content}>
+        <div className={styles.stickers}>
+          <Sticker styleType="neutral" dimension="default">
+            {reward}
+          </Sticker>
+          {claimsLeft !== undefined ? (
+            <Sticker
+              styleType="neutral"
+              dimension="default"
+              withDot={{ dotColor: 'success', dotIcon: DotIcon }}
```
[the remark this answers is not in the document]
flavioislima: Good question. 
I guess it should be red if the claimLeft reach some threshold, but we never defined it.

src/components/RewardsCard/index.module.scss
flavioislima: Hmm, was my first try but since we don't exactly know the size of the images that will come, and in the case mentioned above, the cards won't follow the same size.

flavioislima: image

src/components/RewardsCard/index.module.scss
flavioislima: Same issue as above:
image

flavioislima: Another point actually is that if we make it dynamic we can have cards with different sizes depending on the content. Both on mobile and desktop they have the same size, unless something changed last time I checked there.

src/components/RewardsCard/index.tsx
flavioislima: I wonder if this could be a prop for CardGeneric though.
Because if we keep using it for other types of cards, we will need to add the isLoading on all of them. What do you think?

flavioislima: anyway, added to this one

BrettCleary: yeah it could be a prop for card generic. good point

BrettCleary:
can you fix the test @flavioislima ?

BrettCleary (approved):
lgtm let's just fix the test

src/components/RewardsSection/index.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useCallback, useEffect, useState } from 'react'
+import useEmblaCarousel from 'embla-carousel-react'
+import styles from './RewardsSection.module.scss'
+import RewardsCard, { RewardsCardProps } from '@/components/RewardsCard'
+import ArrowCircularButton from '../ArrowCircularButton'
+
+export interface RewardsSectionProps {
+  rewards: RewardsCardProps[]
+  Link: React.ComponentType<{ children: React.ReactNode; href: string }>
```
BrettCleary, line 9: It would be better to just type this as `any` if this is forcing you to create an entirely new component wrapping `Link` on the store

src/components/RewardsSection/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+  isLoading?: boolean
```
BrettCleary, line 10: this should be able to be set true and the section then shows in a loading skeleton state. 

i.e. we should prepoluate the section with a few dummy cards and have those in a loading state rather than overriding `rewards`.

src/components/RewardsSection/index.tsx
flavioislima: This goes agains what we are doing related with to testing to improve our code quality.
What happens if someone passes an object, number or anything to it?

As you can see on the section, this type is the correct one for accepting a Link from nextJS so I rather keep it otherwise TS makes no point.

src/components/RewardsSection/index.tsx
flavioislima: yeah, this was what I imagined. Will add those cards.

flavioislima: done

flavioislima: So, in the end adding the dummy cards here adds a lot of edge cases and one of them I could not fix on the store:
Ember Sword: there is a quest for it, but no rewards. This makes the section shows but loading forever.
I think might be safer to use a similar approach you did before, but I will add the dummy data on another file.

flavioislima: Actually, even moving the dummy data to the store embers sword still loads forever because this is true:
`props.quests.length > 0 `.

flavioislima: I think that since this is an edge case, I will add a timeout for the isloading state and then we will have layout shift but just for a couple of games. What do you think@

PR #472, opened 2025-03-27
[TECH] Move to PNPM

flavioislima:
One more step to move 100% to PNPM.

Also, Storybook stopped supporting Yarn, although it still works like it says here:

image


BrettCleary (approved):
LGTM

PR #473, opened 2025-03-28
[Feat] Enhance SearchBar with customizable item rendering and styles

eliobricenov:
- Add `renderItem` prop to allow custom rendering of suggestions.
- Introduce `dropdownList` style property for better styling control.
- Update stories to demonstrate new rendering capabilities with images and names.

### DIFF
--- a/src/components/SearchBar/index.tsx
+++ b/src/components/SearchBar/index.tsx
@@ -20,12 +20,15 @@ interface Props extends PopoverProps {
     dropdown?: string
     arrow?: string
     container?: string
+    dropdownList?: string
   }
   styles?: {
     dropdown?: CSSProperties
     arrow?: CSSProperties
     container?: CSSProperties
+    dropdownList?: CSSProperties
   }
+  renderItem?: (suggestion: string) => React.ReactNode
 }
 
 export default function SearchBar({
@@ -37,6 +40,7 @@ export default function SearchBar({
   inputProps,
   classNames,
   styles,
+  renderItem,
   ...props
 }: Props) {
   const input = useRef<HTMLInputElement>(null)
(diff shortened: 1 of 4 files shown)


jiyuu-jin (approved):
So sick, great!  🚀 🌔 

image

flavioislima (approved):
Code looks good. Can you write some tests for the stories as well?

src/components/SearchBar/index.tsx
```diff
@@ -37,6 +40,7 @@ export default function SearchBar({
   inputProps,
   classNames,
   styles,
+  renderItem,
```
BrettCleary, line 43: could we add a default initialization to `(el) => el` so the previous behavior is preserved by default

src/components/SearchBar/index.tsx
```diff
@@ -20,12 +20,15 @@ interface Props extends PopoverProps {
...
     container?: string
+    dropdownList?: string
   }
   styles?: {
     dropdown?: CSSProperties
     arrow?: CSSProperties
     container?: CSSProperties
+    dropdownList?: CSSProperties
   }
+  renderItem?: (suggestion: string) => React.ReactNode
```
BrettCleary, line 31: hmm this pattern seems off to me 
Could we pass in just a React.ReactNode and pass the el as children for a more stylistically react way to decouple?

src/components/SearchBar/index.tsx
eliobricenov: We don't need to. This function is optional, if not present the behavior is the normal one rendering the text.

src/components/SearchBar/index.tsx
eliobricenov: It's a [common react pattern](https://jeffsegovia.dev/blogs/react-render-prop-pattern) but I agree it looks off because we don't use it. I replaced it with your suggestion: https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/473/commits/32a7d55b1bd4534880244cacef431e305c394046

BrettCleary (approved):
lgtm

PR #482, opened 2025-04-07
[Fix] Game info v2 fixes

BrettCleary:
# Summary

This PR:
- adds `BlockchainIcon` with overrides for common chains (hp/chains had some low quality images)
- Add tooltip on hover for blockchain icons in stack
- Add download size sticker
- Add space for editor choice 
- Add null checks for other stickers 
- Update secondary icon button border and fill on hover


biliesilva (approved):
omg I forgot to put null on all stickers 😭

src/components/GameInfoV2/index.tsx
```diff
@@ -119,13 +134,15 @@ const GameInfoV2: React.FC<GameInfoV2Props> = ({
...
                     </Sticker>
                   ) : null,
-                  <Sticker
-                    key={i18n.developer}
-                    styleType="neutral"
-                    variant="filledStrong"
-                  >
-                    {info.developer}
-                  </Sticker>,
+                  info.developer ? (
```
flavioislima, line 137: Same here, it is possible for it to not have developer?

src/components/BlockchainIcon/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { chainIconsSrcOverrides } from '@/utils/chainIconSrcOverrides'
+import BlockchainIconImg from '@/assets/images/BlockchainIconImg.svg?url'
+import { chainMap } from '@hyperplay/chains'
+
+interface BlockchainIconProps {
+  chainId: string
+  className?: string
+  ipfsBaseUrl?: string
+  /* eslint-disable-next-line */
+  ImageComponent?: any // Allow for a custom image component
```
flavioislima, line 11: I think if we start using any we shoudl either disable the not `any` rule for good or start typing properly these components. Otherwise there is no porpuse on using TS only for the basic types 😝 

Another point is that, since we are tripling down on tests in general, even though TS won't replace them, it is a important piece of making a high quality code since we will not pass a number or an object here and break the component, `any` will allow that.

This could be something like this:
`ImageComponent?: FC<ImgHTMLAttributes<HTMLImageElement>>`

flavioislima (changes requested):
Looks mostly good but I think we should make some effort to type everything correctly  and try to achieve clean code for future maintainers and ourselves.

src/components/BlockchainsStack/BlockchainsStack.module.scss
```diff
@@ -142,3 +149,15 @@
   font-size: var(--text-md);
```
biliesilva, line 149: Not sure if this piece here is related to this is, but, as we now have the new typography implemented, we could use eyebrow-sm here so the font doesn't surpass the space inside the circle when there are 2 numbers: 
image

biliesilva: The way it is now, it looks like it is not centered:
image

src/components/BlockchainsStack/index.tsx
```diff
@@ -71,14 +56,20 @@ const BlockchainsStack = ({
...
-              alt={blockchain.name}
-              className={styles.blockchainSvg}
-              onError={() => handleImageError(blockchain.iconUrl)}
-            />
-          </div>
+          <Tooltip
+            key={`blockchain-${index}`}
+            label={blockchain.name}
+            className={styles.tooltip}
+          >
```
biliesilva, line 63: I don't know if Pimentel changed her mind, but when I asked her about the tooltips, we would only have it in the count item:
image

src/components/BlockchainsStack/BlockchainsStack.module.scss
```diff
@@ -142,3 +149,15 @@
...
   font-weight: var(--medium);
 }
+
+.tooltip {
+  max-width: 305px;
+  white-space: normal;
+  overflow-wrap: break-word;
+  border-radius: var(--space-sm-fixed);
+  padding: var(--space-2lg-fixed);
```
biliesilva, line 158: If we want to keep the tooltip, it would be better to decrease the padding to match the size of our other tooltips

biliesilva (changes requested):

src/components/BlockchainsStack/index.tsx
biliesilva, line 63: I don't know if Pimentel changed her mind, but when I asked her about the tooltips, we would only have it in the count item:
image

biliesilva (changes requested):

src/components/GameInfoV2/index.tsx
```diff
@@ -103,13 +116,15 @@ const GameInfoV2: React.FC<GameInfoV2Props> = ({
...
                 title=""
                 items={[
-                  <Sticker
-                    key={i18n.version}
-                    styleType="neutral"
-                    variant="filledStrong"
-                  >
-                    {version}
-                  </Sticker>,
+                  version ? (
```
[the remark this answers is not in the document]
BrettCleary: not sure but this comes from the api so we should be able to gracefully handle undefined or null

src/components/GameInfoV2/index.tsx
BrettCleary: same answer as above

src/utils/chainIconSrcOverrides.ts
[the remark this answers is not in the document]
BrettCleary: We have https://github.com/HyperPlay-Gaming/chains but some of the icons there don't look very good (e.g. have dark or light icons with transparency so we can't show all with the same bg color).

@pimentel08 requested we override the current chain icons for the store chains with the provided icons. We still fall back to chains if the icon is not in this overrides list. 

It does introduce some maintenance. I can compare these icons to what is in chains to see if I can remove some of these overrides for chains that have decent looking icons there.

src/components/BlockchainIcon/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+const BlockchainIcon: React.FC<BlockchainIconProps> = ({
+  chainId,
+  className,
+  ipfsBaseUrl = 'https://ipfs.io/ipfs/',
+  ImageComponent = 'img' // Default to 'img' if no custom component is provided
+}) => {
+  const [missingIcon, setMissingIcon] = useState<boolean>(false)
+
+  let iconUrl = undefined
```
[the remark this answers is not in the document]
BrettCleary: sure thing. just broke it out into a separate hook and pure TS function

src/components/BlockchainIcon/index.tsx
BrettCleary: This is not worth our time and isn't very error prone. Mantine uses `component: any` for this as well. Switching the whole project to JS because of 1 `any` is not logical.

src/components/BlockchainsStack/index.tsx
BrettCleary: She changed her mind during the user testing

src/components/BlockchainsStack/BlockchainsStack.module.scss
BrettCleary: nice catch

src/utils/chainIconSrcOverrides.ts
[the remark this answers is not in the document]
BrettCleary: I was able to remove 6 of the icon overrides. The others either
- don't have ipfs pinned images on ethereum lists
- the icons are in light mode which contrasts with our dark mode theme
- the icon images are dark and have transparency so the icon is not visible
- the icon images are distorted in some other manner

One option is to send upstream pr's with ipfs urls for the missing images and/or polyfill with our desired images in hp/chains. I'll add a TODO for this in this file.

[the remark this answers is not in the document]
BrettCleary: Either way we'll have to maintain updating the icon though

src/components/BlockchainIcon/index.tsx
flavioislima: But that type could work though 
Better than any :P

flavioislima (approved):
LGTM

PR #483, opened 2025-04-10
[FIX] Gamecard size on mobile

flavioislima:
Fix gamecard sizes to fit the mobile grid after fixing the bleeding issue we had before.
Store fix:
https://github.com/HyperPlay-Gaming/hyperplay-store/pull/316

This pull request includes several updates to the `@hyperplay/ui` package, focusing on versioning and styling adjustments for the `GameCard` component.

### Versioning:
* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L3-R3): Incremented the package version from `1.17.0` to `1.17.1`.

### Styling adjustments:
* `src/components/GameCard/GameCard.module.scss`: 
  * Modified the dimensions of the `.root` class to reduce the height from `309px` to `278px` and the width from `200px` to `180px`.
  * Added a media query mixin for mobile devices to adjust the font size of elements within the `.root` class to `var(--text-md)`.


src/components/GameCard/GameCard.module.scss
```diff
@@ -123,6 +123,10 @@ $hoverBorderThickness: 2px;
   @include visibleElement();
   transition: 0.1s;
   position: relative;
+
+  @include for-mobile-only {
+    font-size: var(--text-md) !important;
```
flavioislima, line 128: Needed to add !important here because the title-mixin from the topography is overriding this variable and always setting the title to --text-xl.

Since I don't want to mess with the typography I think its better to add important here.

src/components/GameCard/GameCard.module.scss
```diff
@@ -14,8 +14,8 @@ $hoverBorderThickness: 2px;
 .root {
   box-sizing: border-box;
   position: relative;
-  height: 309px;
-  width: 200px;
+  height: 278px;
+  width: 180px;
```
BrettCleary, line 18: there should be a comment here explaining how it affects the store grid

src/components/GameCard/GameCard.module.scss
BrettCleary: we usually try to stick to the typography types. Title already resizes a bit at some breakpoints is that not enough?

BrettCleary (commented):
Can you check with design first? This is used in the library and I'm not sure we want to change the size everywhere just for so it looks better for mobile on the store

src/components/GameCard/GameCard.module.scss
flavioislima: Yeah, I don't think the breakpoints are working for the gamecard titles (from the gameinfov2 :
image

I discussed this with david here:
https://github.com/HyperPlay-Gaming/px-pm/issues/238

flavioislima:
> Could we change the game name font size to title-small? as we changed Typography and also added a title-sm, we could adjust this to not have the game name so big.

he problem is that title-sm uses --text-xl and I need --text-md. text-xl is too big for the gamecard title on mobile:

```
$titleSmFontSize: --text-xl;
.title-sm {
  font-family: var(--primary-font-family);
  font-size: var($titleSmFontSize);
  font-weight: var(--semibold);
  line-height: 1.5;
  letter-spacing: 0%;
}

biliesilva (approved):

biliesilva:
> > Could we change the game name font size to title-small? as we changed Typography and also added a title-sm, we could adjust this to not have the game name so big.
> 
> he problem is that title-sm uses --text-xl and I need --text-md. text-xl is too big for the gamecard title on mobile:
> 
> ```
> $titleSmFontSize: --text-xl;
> .title-sm {
>   font-family: var(--primary-font-family);
>   font-size: var($titleSmFontSize);
>   font-weight: var(--semibold);
>   line-height: 1.5;
>   letter-spacing: 0%;
> }
> ```

Have you tried title-sm?

BrettCleary (approved):
thx for adding the comments. lgtm!

PR #486, opened 2025-04-11
[FIx] Reward Card and Sections

flavioislima:
Fix stylings after Design Review.
Store PR: https://github.com/HyperPlay-Gaming/hyperplay-store/pull/319

### DIFF
--- a/src/components/RewardsCard/index.tsx
+++ b/src/components/RewardsCard/index.tsx
@@ -13,6 +13,7 @@ export type RewardsCardProps = {
   id: number
   questId: number
   rewardImage: string
+  rewardType: string
   decimals?: number | null
   claimsLeft?: string | null
   rewardName: string
@@ -25,8 +26,9 @@ export type RewardsCardProps = {
 
 function RewardsCard({
   rewardImage,
+  rewardType,
   claimsLeft,
-  decimals,
+  decimals = 0,
   i18n = {
     claimsLabel: 'Claims left'
   },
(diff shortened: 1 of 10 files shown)


src/components/RewardsCard/index.tsx
```diff
@@ -25,22 +26,24 @@ export type RewardsCardProps = {
 
 function RewardsCard({
   rewardImage,
+  rewardType,
   claimsLeft,
-  decimals,
+  decimals = 0,
```
BrettCleary, line 31: this doesn't handle the null case, only the undefined case, so if decimals is null then null is passed to `getDecimalNumberFromAmount`. getDecimalNumberFromAmount doesn't handle the null case

BrettCleary: jsfiddle
image

src/components/RewardsCard/index.tsx
```diff
@@ -25,22 +26,24 @@ export type RewardsCardProps = {
...
-  let rewardText = undefined
+  let rewardText = rewardName
+
   if (amountPerUser) {
     let numToClaim = undefined
     if (amountPerUser && decimals !== undefined && decimals !== null) {
       numToClaim = getDecimalNumberFromAmount(
         amountPerUser,
-        decimals ?? 0
+        decimals
```
BrettCleary, line 46:
```suggestion
        decimals ?? 0
```

BrettCleary (approved):

src/components/RewardsCard/index.tsx
flavioislima: Good catch

src/components/RewardsCard/index.tsx
flavioislima: Yeah, it was like this but then it was breaking all the calcs and returning NaN. I will check how to fix it.

PR #487, opened 2025-04-11
feat(quest): add support for end date

eliobricenov:


src/components/QuestDetails/index.tsx
```diff
@@ -140,6 +149,11 @@ export default function QuestDetails({
     )
   }
 
+  const isEndDateInFuture = endDate && new Date(endDate) > new Date()
+  const utcFormattedDate = dayjs(endDate)
+    .utc()
+    .format('MM/DD [at] HH:mm [(UTC)]')
```
BrettCleary, line 155: hmm why do we want to show this date time in UTC? shouldn't we show it in the user's timezone @pimentel08 @iamdavid0091 ?

BrettCleary: also can we parse into a string format that fits the user's locale? most countries use DD/MM (day/month) date format but US is MM/DD

src/components/QuestDetails/QuestDetails.stories.tsx
```diff
@@ -450,3 +450,17 @@ export const WithExternalSyncButton: Story = {
     ]
   }
 }
+
+export const EndDateInFuture: Story = {
+  args: {
+    ...props,
+    endDate: '2025-04-10T00:00:00.000Z'
```
BrettCleary, line 457: can you move this like 20 years in the future? we are already past this date time so it shows as "ended on"

src/components/QuestDetails/index.module.scss
```diff
@@ -60,3 +60,23 @@
...
+.endDate {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  gap: var(--space-2xs);
+  color: var(--color-neutral-100);
+}
+
+.futureDate {
+  color: var(--color-primary-200);
```
BrettCleary, line 73: can you add a semantic css variable `--text-brand` here https://github.com/HyperPlay-Gaming/hyperplay-ui/blob/607e8abddac48bc89c460c62e17c4063f8da8cf8/src/styles/designSystem/_colors.scss#L106-L122

Also correct me if I'm wrong, but I believe it is `--color-primary-300` on figma https://www.figma.com/design/yHrBPoE70k1I2icW4paEit/Quests?node-id=1356-85942&t=Po4c1ophjvXFfYm5-4

BrettCleary (changes requested):
nice work. requesting some small changes

src/components/QuestDetails/index.tsx
eliobricenov: great call, I found out we can use dayjs for this https://day.js.org/docs/en/display/format#localized-formats

I decided to use the `lll` format.


image

eliobricenov: Design agreed with using the localized `lll` format in the users timezone

BrettCleary (approved):
nice

PR #489, opened 2025-04-11
[Chore] update buttons to match new design

biliesilva:
### Button Component
- Updated Button stories to use Storybook's new CSF format
- Added hover test with proper delays for visual feedback
- Added new button types: 'brand', 'danger-primary', 'danger-secondary', 'danger-tertiary', 'neutral-primary',  'neutral-secondary',  'neutral-tertiary' 
- Added new props: 'focusVisible' and 'hover'
- updated new colors and sizes following the DS on Figma, using our new color variables and typography for button. 

### Typography
- Fix title font size from 2xl to xl

## Some blockers
1 - It was not possible to add the exact hover color as in Figma. Figma mixes the background color with the fill-hover variable, and this becomes like another color. In the code, it overlaps with the background variable color + the --fill-hover variable, which has some transparency; this is particularly difficult with Brand buttons because of the gradient. 
I tested some different styles, but it doesn't work well, so I decided to keep the values ​​for the hover the same way we have done for consistency in the code. 
Also, the hover effect is being applied over the text inside the button too, the same way we do now. I tried to select only the background to apply this state, but it doesn't work, as the text is inside the parent button. I spent a long time trying this without success, so I preferred to keep it as we have today and focus my time on finishing the task. One solution to this would be to rewrite this part in the index and not use a button tag, but a div tag instead and separate the text so that the hover is applied only to the background. But this would not be good for accessibility. If you guys have another solution, I would appreciate suggestions. 

3 - I also couldn't change the text color of the disabled state of the button "primary neutral". It somehow doesn't accept the value of the --text-disabled-inverse color variable; even though this color variable works, it only accepted --text-disabled. Let me know if you guys find where the issues could be.
image

### Design
https://www.figma.com/design/0XLBUyoLZ9m5XVQvlm6Hls/HP-Mantine-UI?node-id=3433-80097&t=QcCXgrBbMaFTv1b0-1


biliesilva:
> > I also couldn't change the text color of the disabled state of the button "primary neutral". It somehow doesn't accept the value of the --text-disabled-inverse color variable; even though this color variable works
> 
> I don't see `--text-disabled-inverse` defined in `src/styles/designSystem/_colors.scss` @biliesilva

I mean --text-inverse-weak 🥲 still confused with a lot of variable names lol

src/components/Button/index.tsx
```diff
@@ -71,9 +85,9 @@ const Button = forwardRef<HTMLButtonElement, ButtonProps>(function Button(
         className={styles.content}
         style={{ gap: `var(--space-${spacing}-fixed)` }}
       >
-        {leftIcon}
-        {children}
-        {rightIcon}
+        {leftIcon && <span className={styles.icon}>{leftIcon}</span>}
+        <span className={styles.text}>{children}</span>
+        {rightIcon && <span className={styles.icon}>{rightIcon}</span>}
```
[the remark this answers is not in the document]
biliesilva: This was when I was trying to make the hover not be applied to the text. I will remove it.

src/components/Button/Button.module.scss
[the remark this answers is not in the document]
biliesilva: > I don't see a hover state for the alert button
> 
> image

We don't have alert button in the new designs, so not sure if make sense to ask @pimentel08 to create variants for this one.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hyperplay/ui",
-  "version": "1.19.2",
+  "version": "1.19.3",
```
flavioislima, line 3: Maybe bump to a minor version since this is a improvement, not a fix.

```suggestion
  "version": "1.20.0",
```

flavioislima (approved):
code LGTM

src/components/Button/Button.module.scss
```diff
@@ -19,7 +22,7 @@ $colorDirection: to top;
   &:disabled,
   &[disabled] {
     cursor: not-allowed;
-    opacity: 0.4;
+    opacity: 3;
```
BrettCleary, line 25: wait why is this 3? isn't 1 the highest opacity can go?

src/components/Button/Button.module.scss
BrettCleary, line 1: image

secondary pressed text color should be `--text-brand`

BrettCleary: same for tertiary

BrettCleary: text color for danger pressed should be `--text-inverse-strong`
image

BrettCleary: danger secondary and danger tertiary pressed text color should be `--text-error`

BrettCleary: we lost the underline animation for link on hover 
image

previous behavior is here https://hyperplay-hjbvl4jjv-hyperplay.vercel.app/?path=/docs/button--docs#large-link

BrettCleary (approved):
looks good. could you fix these issues before merging?

PR #496, opened 2025-04-16
[Chore] update alert cards

biliesilva:
# AlertCard Component Improvements

## Changes Made
- Added new Alert Cards variants based on the new design requirements

## NOTE
The buttons are different from the design because the new buttons have not yet been merged. The[ pr is open here](https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/489) and is under review, and I am currently making the requested changes, but they follow the same state and colors as the designs.

## Design
https://www.figma.com/design/0XLBUyoLZ9m5XVQvlm6Hls/HP-Mantine-UI?node-id=5871-30120&t=cYm6GSLaNMsAgyr9-1


src/components/AlertCard/index.tsx
```diff
@@ -1,78 +1,160 @@
...
+  | 'information'
+  | 'neutral'
+  | 'brand'
+type Size = 'small' | 'large'
+type Layout = 'horizontal' | 'vertical'
 
-export interface InfoAlertProps extends HTMLProps<HTMLDivElement> {
+export interface InfoAlertProps {
   title: string
   message: string
```
flavioislima, line 21: We need to change this `message` to accept a ReactNode otherwise it is not possible to style the text inside it. For instance, if we want to add parts of the string to be bold, italic, etc.

```suggestion
  message: React.ReactNode
```

I added this on this PR due to the ThirdParty downloader alert I need to add to the client, but you can do it on this PR and then I can close mine:
https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/497/files

src/components/AlertCard/index.tsx
```diff
@@ -1,78 +1,160 @@
...
+        </button>
+      )}
+      {showIcon && <div className={styles.icon}>{icon}</div>}
+      <div className={styles.content}>
+        <div className={styles.header}>
+          <div className={styles.title}>{title}</div>
         </div>
+        <div className={styles.message}>{message}</div>
+
+        {listItems && (
```
flavioislima, line 104: I think if `listItems` is a number 0 this won't work as expected.
Better to change it to use a ternary and return `null` if `listItems` is `falsy`.

flavioislima (changes requested):
When testing on storybook I noticed that changing the size prop doesn't do anything. Maybe the styling is not being applied to it?

biliesilva:
> When testing on storybook I noticed that changing the size prop doesn't do anything. Maybe the styling is not being applied to it?

The difference is the size of the description (check the design) it have more than one line on large.

flavioislima:
> > When testing on storybook I noticed that changing the size prop doesn't do anything. Maybe the styling is not being applied to it?
> 
> The difference is the size of the description (check the design) it have more than one line on large.

ah ok, so only for really long descriptions it applies the ellipsis at the end. Was able to check now on storybook.
Maybe add a new story or modify one for Long descriptions so we can understand what this props does.

flavioislima (approved):
Good stuff

src/components/AlertCard/index.tsx
```diff
@@ -1,78 +1,159 @@
...
+              <Button
+                type="tertiary"
+                size="small"
+                onClick={buttons.tertiary.onClick}
+              >
+                {buttons.tertiary.text}
+              </Button>
+            )}
+          </div>
+        ) : null}
```
BrettCleary, line 155: could we just make this a `React.ReactNode` and let the consuming code pass in whatever buttons they want in a `<></>`? Otherwise we'll have to do some prop drilling here if we want to pass in props on the button component like you are doing here with onClick

src/components/AlertCard/index.tsx
```diff
@@ -1,78 +1,159 @@
...
+            <Button
+              type="tertiary"
+              size="small"
+              onClick={link.onClick}
+              rightIcon={<RightArrow />}
+            >
+              {link.text}
+            </Button>
+          </div>
+        ) : null}
```
BrettCleary, line 123: I think you can just cover this use case if you convert the buttonGroup children to a `React.ReactNode` and the consuming code can pass in a link button

src/components/AlertCard/index.tsx
```diff
@@ -1,78 +1,159 @@
...
-      )}
+        <div className={styles.message}>{message}</div>
+
+        {listItems ? (
+          <ul className={styles.list}>
+            {listItems.map((item, index) => (
+              <li key={index}>{item}</li>
+            ))}
+          </ul>
+        ) : null}
```
BrettCleary, line 110: Read this comment after you read my two comments below.

I think this can also be passed it as the button group `React.ReactNode`

src/components/AlertCard/index.tsx
BrettCleary, line 1: warning icon is missing !. probably something with the color setting
image

src/assets/images/CheckmarkCircleOutline.svg
```diff
@@ -1,4 +1,4 @@
-<svg width="16" height="16" viewBox="0 0 16 16" fill="current" xmlns="http://www.w3.org/2000/svg">
+<svg fill="currentColor" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
```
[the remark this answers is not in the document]
BrettCleary: let me know what you think. this also applies to the other svg's listed here.

BrettCleary (commented):
looking good. left a couple comments about previous patterns we have in the repo

src/assets/images/CheckmarkCircleOutline.svg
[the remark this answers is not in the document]
biliesilva: > if you think we should do this differently, I'm open-minded though. It is a different pattern though so we'll need a solid reason to change it

One icon, I don't remember what one, when I removed the view box I could not resize it in the styles even using !important. I deleted the width and height but couldn't delete view box.

src/components/AlertCard/index.tsx
biliesilva: I had a doubt here, because the buttons here in the alert card must be small, and they must have a standard position, where the primary, secondary and link buttons are (if left or right for example. At least that's what I understood from the ui design and from what I understand about the order of buttons, where the primary, secondary and tertiary buttons should be, but since I'm not working on the design anymore, I think we should clear this doubt with @pimentel.

src/components/AlertCard/index.tsx
biliesilva: This a different icon called triangule, same we have in the design. Check the neutrals alert cards.

src/assets/images/CheckmarkCircleOutline.svg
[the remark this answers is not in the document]
BrettCleary: I was referring to changing the fill from `current` to `currentColor`. 

Also, I think you can set viewbox to resize too btw. this is what I do on the website

```
                    <DebianIcon
                      className={DownloadsStyles.platformIcon}
                      viewBox="0 0 32 32"
                    />
```

src/assets/images/AlertOctagon.svg
```diff
@@ -0,0 +1,3 @@
...
...8 2 15.8441 2.02763 15.9592C2.05213 16.0613 2.09253 16.1588 2.14736 16.2483C2.2092 16.3492 2.29568 16.4357 2.46863 16.6086L7.39137 21.5314C7.56432 21.7043 7.6508 21.7908 7.75172 21.8526C7.84119 21.9075 7.93873 21.9479 8.04077 21.9724C8.15586 22 8.27815 22 8.52274 22H15.4773C15.7218 22 15.8441 22 15.9592 21.9724C16.0613 21.9479 16.1588 21.9075 16.2483 21.8526C16.3492 21.7908 16.4357 21.7043 16.6086 21.5314L21.5314 16.6086C21.7043 16.4357 21.7908 16.3492 21.8526 16.2483C21.9075 16.1588 21.9479 16.0613 21.9724 15.9592C22 15.8441 22 15.7218 22 15.4773V8.52274C22 8.27815 22 8.15586 21.9724 8.04077C21.9479 7.93873 21.9075 7.84119 21.8526 7.75172C21.7908 7.6508 21.7043 7.56432 21.5314 7.39137L16.6086 2.46863C16.4357 2.29568 16.3492 2.2092 16.2483 2.14736C16.1588 2.09253 16.0613 2.05213 15.9592 2.02763C15.8441 2 15.7218 2 15.4773 2H8.52274C8.27815 2 8.15586 2 8.04077 2.02763C7.93873 2.05213 7.84119 2.09253 7.75172 2.14736C7.6508 2.2092 7.56432 2.29568 7.39137 2.46863L2.46863 7.39137C2.29568 7.56432 2.2092 7.6508 2.14736 7.75172C2.09253 7.84119 2.05213 7.93873 2.02763 8.04077C2 8.15586 2 8.27815 2 8.52274Z" stroke="#FB7185" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"/>
```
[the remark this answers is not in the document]
biliesilva: As I'm picking some icons we dont have in our UI lib (I believe this pr we need more 3 icons), when we save in SVG it also comes with a color value or none, and also a size, some of them that I remember I was having problem in styling and that I edited, I removed the width/height and in the fill I put the "currentColor" value.

PR #500, opened 2025-04-21
[Chore] update alert global 

biliesilva:
# Alert Global Updates

## Layout Changes
- Changed border style from full border to bottom border only
- Removed border radius
- Change background color
- Added responsive padding adjustments for mobile

## Container Structure
- Reorganized component structure with nested containers for better layout control
- Added `.content` class for better content organization
- Improved spacing with consistent gap values using design system variables

## Responsive Improvements
- Added mobile breakpoint

## Icon Adjustments
- Refactor icons with the icons from our DS, also adding new colors

## Button and Close Button
- Add Button prop

## General Cleanup
- Removed unnecessary positioning styles
- Simplified margin and padding values
- Improved overall component structure to match our new DS system

[Design](https://www.figma.com/design/0XLBUyoLZ9m5XVQvlm6Hls/HP-Mantine-UI?node-id=5801-57029&t=K6z26vhn0evBTNKF-1)


src/assets/images/AlertTriangle.svg
```diff
@@ -1,3 +1,3 @@
...
+<path d="M7.99976 6.00015V8.66682M7.99976 11.3335H8.00642M7.07663 2.59464L1.59338 12.0657C1.28924 12.591 1.13717 12.8537 1.15965 13.0693C1.17925 13.2573 1.27776 13.4282 1.43067 13.5394C1.60597 13.6668 1.90948 13.6668 2.5165 13.6668H13.483C14.09 13.6668 14.3935 13.6668 14.5688 13.5394C14.7217 13.4282 14.8203 13.2573 14.8399 13.0693C14.8623 12.8537 14.7103 12.591 14.4061 12.0657L8.92288 2.59463C8.61983 2.07119 8.46831 1.80947 8.27062 1.72157C8.09818 1.64489 7.90133 1.64489 7.72889 1.72157C7.5312 1.80947 7.37968 2.07119 7.07663 2.59464Z" stroke="currentColor" stroke-width="1.2" stroke-linecap="round" stroke-linejoin="round"/>
```
BrettCleary, line 2: I thought we decided on `stroke="current"`

BrettCleary: Also do we want 1.2 to be the stroke width? It won't be pixel perfect if we have a fraction here right?

src/components/Alert/index.tsx
```diff
@@ -41,8 +56,31 @@ const Alert = ({
...
       {...props}
     >
-      {Icon && <Icon className={cs(styles.icon, classNames?.icon)} />}
-      <span className={cs(classNames?.message)}>{message}</span>
+      <div className={styles.container}>
+        <div className={styles.content}>
+          <Icon className={cs(styles.icon, classNames?.icon)} />
+          <p className={cs(styles.message, 'eyebrow', classNames?.message)}>
+            {message}
+          </p>
```
BrettCleary, line 64: if message is undefined, I don't think we should render this `p` element

src/components/Alert/index.tsx
```diff
@@ -41,8 +56,31 @@ const Alert = ({
...
+            {button}
+          </Button>
+        ) : null}
+      </div>
+      <CloseButton
+        type="button"
+        className={cs(styles.closeButton, classNames?.closeButton)}
+        aria-label="Close"
+      >
+        {closeButton}
```
BrettCleary, line 82: if closeButton isn't defined or is null, shouldn't we hide this `CloseButton` element?

BrettCleary (approved):
left some small notes. feel free to merge after responding/resolving them :ok_hand:

src/assets/images/AlertTriangle.svg
biliesilva: The current only is not working, some icons, if I change to current, everything turns black or disappears, the color is only applied with the currentColor.

biliesilva: This icon naturally already had a stroke, but it was too thick compared to the others on Alert. I made it a little thinner to maintain consistency.

src/components/Alert/index.tsx
biliesilva: Before I change, the message was not with the? (optional) for the type. I ended up changing it to ? but I will leave it as it was.

src/components/Alert/index.tsx
biliesilva: We do not have a variant without a close button in the designs, all these global alerts have the close button X

BrettCleary: it is optional in the props (`closeButton?: React.ReactNode`) so you should either remove the `?` or handle the optional case

PR #502, opened 2025-04-22
[Fix] fix reward cut off

biliesilva:
# QuestCard Component Updates

Needed here: http://github.com/HyperPlay-Gaming/hyperplay-store/pull/332

## Visual Design Improvements
- Updated color tokens to use semantic design system variables:
  - Improved consistency with the design system
  -  Layout and Spacing Refinements
  - Added ellipsis handling for multiple text elements
  - Description now limited to 1 line with ellipsis


flavioislima (approved):
Tested here and lgtm

src/components/QuestCard/index.module.scss
```diff
@@ -12,17 +12,17 @@ $border: 2px var(--color-neutral-600) solid;
 .image {
   border: $border;
   border-bottom: 0px;
-  max-height: 150px;
+  max-height: 156.75px;
```
BrettCleary, line 15: Could we round this to the nearest whole number for pixel perfection?

src/components/QuestCard/index.module.scss
```diff
@@ -109,9 +108,16 @@ $border: 2px var(--color-neutral-600) solid;
 }
 
 .currencyName {
-  color: var(--color-neutral-300);
-  font-size: var(--text-md);
+  color: var(--text-weak);
   padding-top: var(--space-3xs);
+  overflow: hidden;
+  text-overflow: ellipsis;
```
BrettCleary, line 114: hmm wondering if we should add a tooltip on hover here so users have a way to still see the full reward name? cc @pimentel08

BrettCleary: same question for `.description`

BrettCleary (commented):
Can we double check with design on text overflow edge case implementation?

BrettCleary (approved):

PR #511, opened 2025-05-08
[Fix] secondary gradient button after for client

BrettCleary:
# Summary

Hover effect does not work on client due to `z-index: -1` for the after

### DIFF
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -437,7 +437,6 @@ button.base {
     content: '';
     position: absolute;
     inset: 1px;
-    z-index: -1;
     transition: background-color 0.2s ease;
   }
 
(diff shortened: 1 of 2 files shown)


jiyuu-jin (approved):

eliobricenov (approved):

src/components/Button/Button.module.scss
```diff
@@ -437,7 +437,6 @@ button.base {
     content: '';
     position: absolute;
     inset: 1px;
-    z-index: -1;
```
biliesilva, line 440: Seems like it's working well with this removal. I believe I forgot to remove this line after changing the opacity that was the real problem with the hover being above the text at the time.

biliesilva (approved):

PR #524, opened 2025-06-23
fix: remove invalid css

eliobricenov:


jiyuu-jin (approved):
Yoooo, hell yeahhh, I tried this but I think I must have missed the other svg paths.  So sick bro! 🚀

BrettCleary (approved):
:fire:

PR #530, opened 2025-06-29
[Feat] Updates to options accordion

BrettCleary:
# Summary

- change `setOptions`
- add `displayName` to support chain id's as filter keys with chain name as the display text for filtering

### DIFF
--- a/src/components/OptionsAccordion/OptionsAccordion.stories.tsx
+++ b/src/components/OptionsAccordion/OptionsAccordion.stories.tsx
@@ -2,58 +2,65 @@ import React, { useState } from 'react'
 
 import { StoryObj } from '@storybook/react'
 
-import OptionsAccordion from '.'
+import OptionsAccordion, { OptionsType, PanelOptions } from '.'
 
 export default {
   title: 'OptionsAccordion',
   component: OptionsAccordion
 }
 
-type OptionsType = { [key: string]: boolean }
-const Genre: OptionsType = {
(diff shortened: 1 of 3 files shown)


src/components/OptionsAccordion/OptionsAccordion.stories.tsx
```diff
@@ -2,58 +2,65 @@ import React, { useState } from 'react'
...
 }
 
-const Scrollable: OptionsType = Object.fromEntries(
-  Array.from({ length: 25 }, (_, i) => [`Item${i + 1}`, false])
+const Chains: PanelOptions = {
+  1: { selected: false, displayName: 'Ethereum Mainnet' },
+  2: { selected: false, displayName: 'Some Other Chain' },
+  1010101: { selected: false, displayName: 'Another Chain' },
+  1231123: { selected: false }
+}
```
eliobricenov, line 47: Couldn't you map the chains to their display name and use them as the key? That way we can keep the current API and not need the `selected` property?

BrettCleary: Hmm not sure if I understand. The key I think we need to use is the chainId (and caip 2 https://chainagnostic.org/CAIPs/caip-2 once we have non-evm tokens), but the previous implementation only used the key as the display value.

Are you proposing we just have a separate map prop for chainId -> displayName? I'd be fine with that

PR #532, opened 2025-06-30
Feat/implement masonry layout component for New games section

biliesilva:
Closes: https://github.com/orgs/HyperPlay-Gaming/projects/16/views/8?pane=issue&itemId=100172865&issue=HyperPlay-Gaming%7Chyperplay-pm%7C106

### DIFF
--- a/src/components/MasonryLayout/index.tsx
+++ b/src/components/MasonryLayout/index.tsx
@@ -1,21 +1,13 @@
 import React from 'react'
 import styles from './MasonryLayout.module.scss'
 import gameImage from '@/assets/banners/QuestCardV2Image.png?url'
-import Image from 'next/image'
 
 export interface MasonryLayoutProps {
   imageUrl?: React.ReactNode
 }
 
 const MasonryLayout = ({ imageUrl }: MasonryLayoutProps) => {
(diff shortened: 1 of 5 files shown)


src/components/MasonryLayout/index.tsx
BrettCleary, line 1: this only takes a single `imageUrl` but needs  to support different images for each position in the masonry layout

src/components/MasonryLayout/index.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import styles from './MasonryLayout.module.scss'
+import gameImage from '@/assets/banners/QuestCardV2Image.png?url'
+
+export interface MasonryLayoutProps {
+  imageUrl?: React.ReactNode
```
BrettCleary, line 6: if this is just a react node, we should probably rename it since it isn't a url

BrettCleary (changes requested):

BrettCleary:
@biliesilva you still need to export this component from `src/index.ts`

eliobricenov (approved):

eliobricenov:
great work, looks great!

PR #534, opened 2025-07-01
[Feat] add game card loading state

BrettCleary:
# Summary

Add `isLoading` with shiny loader for game card


BrettCleary:
closing since we'll use a skeleton outside of this component to keep this component from becoming more complicated

PR #536, opened 2025-07-02
Feat/implement horizontal card layout component

biliesilva:
Closes: https://github.com/orgs/HyperPlay-Gaming/projects/16/views/8?pane=issue&itemId=101377048&issue=HyperPlay-Gaming%7Chyperplay-pm%7C139


src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const image = gameImage || (
+    <img src="/MoonBlastersCover.png" alt="Game Image" />
+  )
+
+  return (
+    <>
+      <div
+        className={[styles.horizontalCard, styles[tone], styles[size]].join(
+          ' '
+        )}
```
BrettCleary, line 41: we have a `classNames` package which is very useful for situations like this

```
import classNames from 'classNames'
```

```suggestion
        className={classNames(styles.horizontalCard, styles[tone], styles[size])}
```

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+        onClick={onCardClick}
+      >
+        <div className={styles.gameImage}>{image}</div>
+        <div className={styles.content}>
+          <div className={['title', styles.title].join(' ')}>{title}</div>
```
BrettCleary, line 46: can use classNames here too

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+          {button ? (
+            <div className={styles.button}>
+              <Button type="secondary" size="small" onClick={onButtonClick}>
+                {button}
+              </Button>
+            </div>
+          ) : null}
+        </div>
+        {orderNumber ? (
+          <div className={`h6 ${styles.orderNumber}`}>{orderNumber}</div>
```
BrettCleary, line 61: can use classNames here too

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+              <Sticker>{badge}</Sticker>
+            </div>
+          ) : null}
+          {button ? (
+            <div className={styles.button}>
+              <Button type="secondary" size="small" onClick={onButtonClick}>
+                {button}
+              </Button>
+            </div>
+          ) : null}
```
BrettCleary, line 58: I would honestly just replace all of this with one ReactNode prop

src/components/HorizontalCard/HorizontalCard.module.scss
BrettCleary, line 1: it's overflowing on small mobile. Could we update the styles so it is responsive here and only fills the available width?

image

BrettCleary: small card is overflowing too 
image

BrettCleary:
@biliesilva you still need to export this component from `src/index.ts`

https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/536#discussion_r2183536698

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  badge,
+  button,
+  orderNumber,
+  onCardClick,
+  tone = 'neutral',
+  size = 'large'
+}) => {
+  const image = gameImage || (
+    <img src="/MoonBlastersCover.png" alt="Game Image" />
+  )
```
BrettCleary, line 31: This should be the default fallback image instead of moon blasters

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+          styles[tone],
+          styles[size]
+        )}
+        onClick={onCardClick}
+      >
+        <div className={styles.gameImage}>{image}</div>
+        <div className={styles.content}>
+          <div className={classNames(styles.title, 'title')}>{title}</div>
+          {badge && <div className={styles.badge}>{badge}</div>}
+          {button && <div className={styles.button}>{button}</div>}
```
BrettCleary, line 47: nitpick but could we make these ternaries too like you have below with ` ? : null`

src/components/HorizontalCard/HorizontalCard.module.scss
```diff
@@ -0,0 +1,163 @@
...
+.horizontalCard {
+  display: flex;
+  align-items: center;
+  padding: var(--space-md);
+  cursor: pointer;
+  width: 100%;
+  overflow: hidden;
+
+  @include for-mobile-only {
+    width: 100% !important;
```
BrettCleary, line 12: do we need this `!important`? it will force a recalculation of the stylesheet, hurting our load perfomance

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+
+import styles from './HorizontalCard.module.scss'
+
+import classNames from 'classnames'
+
+export interface HorizontalCardProps {
```
BrettCleary, line 7: same comment as here https://github.com/HyperPlay-Gaming/hyperplay-ui/pull/532#discussion_r2190814222

```suggestion
export interface HorizontalCardProps extends React.HTMLAttributes<HTMLDivElement> {
```

src/components/HorizontalCard/index.tsx
BrettCleary, line 47: I really think we should just take an arbitrary react node for this section instead of only a badge or button.

We could also just pass in `children` here for a nice react styled syntax

I also don't see the benefit of having both here since it just accepts a ReactNode and I don't see `styles.badge` `style.button` defined in the scss module

BrettCleary (changes requested):
requesting changes since we need to change the fallback image

pimentel08 (commented):
The layout is looking great! For the view with only one card visible, could we keep the image size consistent with the two-card layout, rather than expanding the single card to full width as shown in these screenshots?

Screenshot 2025-07-07 at 11 47 28 AM
Screenshot 2025-07-07 at 11 47 34 AM

src/components/HorizontalCard/HorizontalCard.module.scss
biliesilva: I don't know why but the component is not taking up the full width on the small mobile, the !important here also doesn't seem to work in any way.

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+const HorizontalCard: React.FC<HorizontalCardProps> = ({
+  gameImage,
+  title,
+  children,
+  orderNumber,
+  onCardClick,
+  tone = 'neutral',
+  size = 'large'
+}) => {
+  const image = gameImage || <img src="/fallback_card.jpg" alt="Game Image" />
```
BrettCleary, line 28: I think this may break when you pull it into the store unless you import the FallbackImage like we do in the `GameCard`:

```suggestion
import styles from './HorizontalCard.module.scss'

import classNames from 'classnames'
import FallbackImage from '@/assets/fallback_card.jpg?url'

export interface HorizontalCardProps
  extends React.HTMLAttributes<HTMLDivElement> {
  gameImage?: React.JSX.Element
  imageUrl?: string
  children?: React.ReactNode
  title: string
  orderNumber?: string
  onCardClick: React.MouseEventHandler<HTMLDivElement>
  tone?: 'brand' | 'neutral'
  size?: 'large' | 'small'
}

const HorizontalCard: React.FC<HorizontalCardProps> = ({
  gameImage,
  title,
  children,
  orderNumber,
  onCardClick,
  tone = 'neutral',
  size = 'large'
}) => {
  const image = gameImage || <img src={FallbackImage} alt="Game Image" />
```

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const HorizontalCard: React.FC<HorizontalCardProps> = ({
+  gameImage,
+  title,
+  children,
+  orderNumber,
+  onCardClick,
+  tone = 'neutral',
+  size = 'large'
```
BrettCleary, line 26:
```suggestion
  size = 'large',
  ...props
```

src/components/HorizontalCard/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <>
+      <div
+        className={classNames(
+          styles.horizontalCard,
+          styles[tone],
+          styles[size]
+        )}
+        onClick={onCardClick}
```
BrettCleary, line 38: This way we're more un-opinionated and extensible with the component styling

```suggestion
        onClick={onCardClick}
        {...props}
```

src/components/HorizontalCard/HorizontalCard.module.scss
BrettCleary, line 1: I'm seeing an odd issue in the bottom right on hover gradient border for the brand large card where the thickness isn't consistent for the border. 

image

BrettCleary (approved):
Looks great! thanks for making the changes. Let's :ship: it

PR #539, opened 2025-07-09
[Tech] Build ssr compatible module export 

BrettCleary:


package.json
```diff
@@ -75,11 +75,11 @@
     "prettier": "^2.8.3",
     "react": "^19.1.0",
     "react-dom": "^19.1.0",
-    "react-loading-skeleton": "^3.5.0",
```
BrettCleary, line 78: we should be using our shiny loader mixin instead of this, so removing it

package.json
```diff
@@ -54,7 +54,7 @@
     "@typescript-eslint/parser": "^8.18.0",
     "@vitejs/plugin-react-swc": "^3.1.0",
     "babel-loader": "^8.3.0",
-    "classnames": "^2.3.2",
+    "classnames": "^2.5.1",
```
BrettCleary, line 57: the build output for classnames has a window reference, preventing us from removing `"use client"` directive on components like `Button`. Bumped to see if there was a fix but there wasn't. I think the issue has to do with how vite is building it. Maybe we need to mark it as external but I'd rather explore on a separate branch.

rollup-exclude-use-client.ts
```diff
@@ -0,0 +1,8 @@
+// Files names that should not have use client directive at the top of the output file
+export const ROLLUP_EXCLUDE_USE_CLIENT = ['src/index'].reduce<string[]>(
```
BrettCleary, line 2: similar approach to mantine here

src/index.scss
BrettCleary, line 1: we can just use `src/styles/index.scss`. no need for this file

vite.config.ts
```diff
@@ -35,6 +37,16 @@ export default defineConfig({
           dest: ''
         }
       ]
+    }),
+    banner((chunk) => {
+      if (
+        !ROLLUP_EXCLUDE_USE_CLIENT.includes(chunk.fileName) &&
+        !chunk.fileName.includes('node_modules')
+      ) {
+        return "'use client';\n"
```
BrettCleary, line 46: adds use client directive to all of our components by default now

eliobricenov (approved):

PR #545, opened 2025-07-15
Fix/add class name prop and resize min width of horizontal card

biliesilva:
Remove min-width for the large variant and add a className prop for additional style if needed


src/components/HorizontalCard/HorizontalCard.tsx
```diff
@@ -45,6 +47,7 @@ const HorizontalCard = React.forwardRef<HTMLDivElement, HorizontalCardProps>(
       <CardComponent
         ref={ref}
         className={classNames(
+          className,
```
BrettCleary, line 50: I would recommend putting this at the end of these `classNames(...)` args. This way if it's the same specificity as the others class args, it will still override

PR #550, opened 2025-07-23
Enhance Sticker component with tooltip support and text truncation

biliesilva:
Closes: https://github.com/HyperPlay-Gaming/hyperplay-pm/issues/723

I added a width: 100% for the Sticker, which it didn't have before, so when it is inside other components, like for example the HorizontalCard, and this has a width limit, the sticker (badge) will have an ellipsis in the text and will show a tooltip onHover.

https://hyperplay-ui-git-fix-adjust-badges-hyperplay.vercel.app/?path=/story/sticker--with-tooltip

Could not add the build for store visualization because it was generating a lot of errors with mantine, carousel, and other builds, after right I added the hyperplay-ui test version. 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperplay/ui",
-  "version": "1.33.9",
+  "version": "1.33.10",
   "license": "LGPL-3.0-only",
   "scripts": {
     "dev": "vite",


eliobricenov (approved):

src/components/Sticker/index.tsx
```diff
@@ -68,7 +89,30 @@ export default function Sticker({
...
+          opened={showPopover}
+          offset={{ mainAxis: 16 }}
+          styles={popoverStyles}
+        >
+          <Popover.Target>
+            <span
+              className={styles.text}
+              onMouseEnter={open}
+              onMouseLeave={close}
+            >
```
BrettCleary, line 105: did the `Tooltip` component from mantine not work here?

src/components/Sticker/index.tsx
```diff
@@ -68,7 +89,30 @@ export default function Sticker({
             : null}
         </span>
       )}
-      {children}
+      {showTooltip ? (
+        <Popover
```
BrettCleary, line 93: might want to consider moving this to wrap the root div returned by this component. Currently if you hover over the edge of the sticker, it still shows the on hover effect but it does not show the tooltip.

The tooltip is only showing if you hover the text

BrettCleary (approved):

src/components/Sticker/index.tsx
biliesilva: No :(

PR #560, opened 2025-09-10
Feat/quest back button

iabubakardev:


jiyuu-jin:
Looks good, but I think you need to run `pnpm prettier-fix` to fix the linting.

